PR #2, opened 2015-11-16
Various updates to the parser

danielgtaylor:
- Better JSON Schema reference handling (including external sources)
- Do not set resource titles
- Use `summary` instead of `operationId` for action titles
- Set a relation value based on `operationId` for actions
- Update test fixtures to support the above

cc @smizell @netmilk


smizell:
Code looks good to me.

danielgtaylor:
@smizell everything should be up to date now, with the same coverage as before. The only difference is that some covered lines were removed, so the overall coverage went down by a tiny fraction. I think this should be safe to merge.

src/adapter.js
```diff
@@ -117,175 +103,171 @@ export function parse({minim, source}, done) {
...
 
-      // Currently, default responses are not supported in API Description format
-      const relevantResponses = _.omit(methodValue.responses, 'default');
+        const transition = new Transition();
+        resource.content.push(transition);
 
-      if (_.keys(relevantResponses).length === 0) {
-        createTransaction(minim, transition, method);
-      }
+        transition.meta.set('title', methodValue.summary || methodValue.operationId || '');
```
smizell: @netmilk You OK with falling back to `operationId` here? Or would you rather it be blank?

smizell:
@danielgtaylor I had one question about falling back to `operationId`, but I think it is good to merge. We can figure that out later.

src/adapter.js
netmilk: @smizell @danielgtaylor I'm sorry, but the title should be left empty in the refract if there's no summary in swagger. Style Guide engine other dependant tooling wouldn't know it was empty in the swagger.

pksunkara:
:+1:

PR #4, opened 2015-11-18
Add test fixtures for various features

danielgtaylor:
This is a good start to having minimal examples covering all the various Swagger features we support. A non-exhaustive list, including things I haven't done that I think we should do at some point:

* [x] Minimal examples in JSON and YAML
* [x] API title and description
* [x] Operation and action
* [x] Parameters (path, query, and body)
* [x] Path object references
* [x] JSON Schema references
* [x] HTTP headers
* [x] Request parameters without responses
* [x] Base path and host name
* [x] Authentication
* [x] Path-level parameters (@smizell)
* [x] All the different request verb types
* [x] External document references
* [x] Support for tags
* [x] Form parameters
* [x] Default responses (@smizell)
* [x] Swagger [extensions](https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#paths-object) (we should ignore these `x-foo` items)

Also included is the large pet store example, but I'm trying to keep the above fixtures as small as possible.

cc @smizell @netmilk


danielgtaylor:
@smizell this should now be good to go. Give it a look and feel free to merge and release!

src/adapter.js
```diff
@@ -192,10 +317,20 @@ export function parse({minim, source}, done) {
...
+          .omit('default')
+          .omit(isExtension)
+          .value();
 
         if (_.keys(relevantResponses).length === 0) {
-          createTransaction(minim, transition, method);
+          if (bodyParameters.length) {
+            // Create an empty successful response so that the request/response
+            // pair gets properly generated.
+            relevantResponses[200] = {};
```
smizell: My concern here is that we're making some assumption from the document, when I believe other tooling should know what to do when there is no response found.

src/adapter.js
```diff
@@ -118,21 +164,84 @@ export function parse({minim, source}, done) {
...
+
+    if (swagger.info) {
+      if (swagger.info.title) {
+        api.meta.set('title', swagger.info.title);
+      }
+
+      if (swagger.info.description) {
+        api.content.push(new Copy(swagger.info.description));
+      }
+    }
```
smizell: Just a note, you can do this with lodash without testing each step.

```
if (_.get(swagger, 'info.title') { ... }
```

src/adapter.js
```diff
@@ -12,7 +12,51 @@ export const mediaTypes = [
...
+    : source.swagger === '2.0');
+}
+
+// Test whether a key is a special Swagger extension.
+function isExtension(value, key) {
+  return key.indexOf('x-') === 0;
+}
+
+// Test whether tags can be treated as resource groups, and if so it sets a
+// group name for each resource (used later to create groups).
```
smizell: After seeing the implementation of this, I'm wondering if we need to step back and think through this issue. I'm not sure if it's right to partially support tags.

Would be interested in @netmilk's thoughts.

src/adapter.js
smizell: Would be interested in your thoughts @netmilk

smizell:
Reviewed the code and have a couple of concerns with making assumptions.

src/adapter.js
netmilk: @smizell @danielgtaylor The response should be definitely left empty. We will have to write some design rules exactly for this case and we would know there was no responses given in the source swagger.

danielgtaylor: The bug this is trying to solve is that before this change if there were request parameters defined but no responses defined, then we would *not* generate a request refract object and the input parameters would be completely ignored.

We can either make my change above or we can refactor the other code to only create a request rather than a request and an empty response. Given that we already create blank requests all over the place I think my solution above makes sense, and I'm struggling to understand cases where this would be an issue in other tooling.

smizell: @danielgtaylor @netmilk So we include an empty response rather than no response or one with status in it. Responses have no required attributes in the Refract namespace, but a response element is a MUST. Therefore, to satisfy namespace, empty response makes sense to me.

src/adapter.js
danielgtaylor: This feature is important for documentation and organization. All of the real-world examples I have seen use tags in such a way that we can treat them as resource groups. Since refract doesn't support a tagging feature and if/when it gets added it means an extra burden on tooling authors (e.g. Aglio then needs to support **both** resource groups *and* tags) I'm for trying it this way.

If we discover some issues with it then I am okay reevaluating this position, but at the moment (with no user feedback) I believe that users will see this feature as treating Swagger more like a first class citizen in Fury.

src/adapter.js
smizell: Sorry @danielgtaylor your message did not show up before I posted mine.

I think the issue is that it's adding a status code along with the response instead of it being truly empty. Thoughts on that?

src/adapter.js
smizell: I believe this goes back to your comment on how we handle resource groups in Refract. Seems like the best solution is to go with what you have, like you've said, but think through things in the API Description spec itself to have better support for tags.

Thoughts?

netmilk: @smizell @danielgtaylor I would consider Swagger Tags out of scope of the MVP. I would propose to throw a annotation with warning if there are tags present in the swagger object or in the operation object. Does this work for you guys?

netmilk: Swagger tags are overlapping transition groups and I thing we have to come up with support for this in the refract API description namespace.

smizell: @danielgtaylor Implementation looks good to go per our discussion.

PR #6, opened 2015-11-19
Use generic title so same parse result can be used

smizell:


danielgtaylor:
Coveralls is stuck, but this looks fine to me.

PR #7, opened 2015-11-19
Switch to using Swagger Zoo

danielgtaylor:
Most of the diff is just removing the fixtures since they are all in the zoo now.

Edit: It will fail until the zoo module is published. Have it working locally via `npm link`.

cc @smizell


smizell:
Good to go when zoo is published.

pksunkara:
:+1:

PR #12, opened 2015-11-25
Fixes URI Template when no query parameters specified.

freaz:


smizell:
@freaz This looks good. Were there any fixtures that needed to be changed? If not, do you think we should add one?

danielgtaylor:
LGTM :+1:

freaz:
@smizell I didn't find it in feature fixtures. But it is in zoo example for instagram: https://github.com/apiaryio/swagger-zoo/blob/master/fixtures/examples/refract/instagram.json#L99

smizell:
@freaz I see. That Instagram example isn't in the adapter fixtures. 

Also note, we usually try to add unpublished changes to the changelog file. Would you mind adding a note there with a note about the bugfix? Next would be to open a PR to bump version number and publish.

freaz:
@smizell Yes sure.

PR #14, opened 2015-11-25
Removes extra slash in basePath

freaz:


smizell:
One concern is that in correcting this, you may create another case where people do as follows:

```yaml
swagger: '2.0'
info:
  title: Swagger with extra slashes
basePath: '/api/'
paths:
  'resource':
    get:
      200:
        description: 'Success'
```

Do you think we should think about this, too?

smizell:
Also, please remember to add this new fixture to Swagger Zoo :(

freaz:
@smizell I think it's not possible defined only `resource` https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#paths-object.

Thanks for reminding zoo.

smizell:
@freaz Great, thanks.

PR #16, opened 2015-11-26
Implement source maps

danielgtaylor:
This implements source maps for a handful of elements and metadata properties. Here is how loading works when `generateSourceMap: true` is passed as an option:

1. Load YAML input data
2. Load YAML input again, this time into an AST
3. Dereference the loaded data (*not* AST)
4. Begin traversing loaded dereferenced data to create Refract elements
5. Anytime we need a source map, traverse the AST (handling references) until the node we care about is found, then get the start/end location from the original file.
6. Done!

See the test file for example input/output.

Note: tests will fail until https://github.com/refractproject/minim/pull/89 is resolved.

cc @smizell

### DIFF
--- a/src/adapter.js
+++ b/src/adapter.js
@@ -1,7 +1,8 @@
 import _ from 'underscore';
+import buildUriTemplate from './uri-template';
 import $RefParser from 'json-schema-ref-parser';
 import yaml from 'js-yaml';
-import buildUriTemplate from './uri-template';
+import yamlAst from 'yaml-js';
 
 export const name = 'swagger';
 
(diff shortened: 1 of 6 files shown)


src/adapter.js
```diff
@@ -59,6 +60,84 @@ function useResourceGroups(api) {
   return tags.length > 0;
 }
 
+// Look up a position in the original source based on a JSON path, for
+// example 'paths./test.get.responses.200'
+function getPosition(ast, path) {
```
danielgtaylor: The input JSON path should probably support some kind of escaping mechanism. Maybe just percent-encode the periods and square brackets? Ideas?

src/adapter.js
```diff
@@ -168,10 +255,18 @@ export function parse({minim, source}, done) {
     if (swagger.info) {
       if (swagger.info.title) {
         api.meta.set('title', swagger.info.title);
+
+        if (generateSourceMap && ast) {
```
smizell: Could you simply use `if (ast)`?

src/adapter.js
smizell: If it is true [JSONPath](http://goessner.net/articles/JsonPath/), you should be able to wrap keys in brackets.

```
x.store.book[0].title
```

is the same as:

```
x['store']['book'][0]['title']
```

smizell:
Makes total sense. Great work.

src/adapter.js
danielgtaylor: Good point. I'll look into this a bit.

PR #20, opened 2015-12-01
Implement annotations

danielgtaylor:
This PR implements adding annotations to the parse result. It covers all `TODO` items marked for a future annotation and adds a new unit test for all of them.

cc @smizell

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,3 +28,20 @@ fury.parse({source: '... your Swagger 2.0 document ...'}, (err, result) => {
   console.log(result.api.title);
 });
 ```
+
+### Parser Codes
+
+The following codes are used by the parser when creating warning and error annotations.
+
+Warnings:
+
+Code | Description
+---: | -----------
+   2 | Source maps are unavailable due either to the input format or an issue parsing the input.
+   3 | Data is being lost in the conversion.
+
+Errors:
+
+Code | Description
+---: | -----------
+   1 | Error parsing input (e.g. malformed YAML).
--- a/src/adapter.js
+++ b/src/adapter.js
@@ -4,6 +4,24 @@ import $RefParser from 'json-schema-ref-parser';
 import yaml from 'js-yaml';
 import yamlAst from 'yaml-js';
 
+// These describe the type of annotations that are produced by this parser
+// and assigns a unique code to each one. Downstream applications can use this
+// code to group similar types of annotations together.
+const ANNOTATIONS = {
+  CANNOT_PARSE: {
+    type: 'error',
+    code: 1,
+  },
+  AST_UNAVAILABLE: {
+    type: 'warning',
+    code: 2,
+  },
+  DATA_LOST: {
+    type: 'warning',
+    code: 3,
+  },
+};
+
 export const name = 'swagger';
 
 // TODO: Figure out media type for Swagger 2.0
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -28,3 +28,20 @@ fury.parse({source: '... your Swagger 2.0 document ...'}, (err, result) => {
...
+Code | Description
+---: | -----------
+   2 | Source maps are unavailable due either to the input format or an issue parsing the input.
+   3 | Data is being lost in the conversion.
+
+Errors:
+
+Code | Description
+---: | -----------
+   1 | Error parsing input (e.g. malformed YAML).
```
smizell: I think this is more of a design comment than in context of this PR, but I'd love to see this kind of thing solved again with hypermedia. For instance, instead of having numbers that correspond to errors, we have a URL where errors are defined and we link to these errors. This is a way of decoupling from the actual implementation.

Just something to think about for the future.

src/adapter.js
```diff
@@ -4,6 +4,24 @@ import $RefParser from 'json-schema-ref-parser';
 import yaml from 'js-yaml';
 import yamlAst from 'yaml-js';
 
+// These describe the type of annotations that are produced by this parser
+// and assigns a unique code to each one. Downstream applications can use this
+// code to group similar types of annotations together.
+const ANNOTATIONS = {
+  CANNOT_PARSE: {
+    type: 'error',
+    code: 1,
```
smizell: Per my comment about using hyperlinks for this, `type` could be a link relation and `code` could be replaced by a URL.

src/adapter.js
```diff
@@ -377,7 +444,9 @@ export function parse({minim, source, generateSourceMap}, done) {
         }
 
         if (methodValue.externalDocs) {
-          // TODO: [Annotation] Warn about unused external docs.
+          setupAnnotation(ANNOTATIONS.DATA_LOST,
+            `paths.${href}.${method}.externalDocs`,
+            'External documentation is not yet supported');
```
smizell: Another thought for the future, this could be solved by adding a link.

src/adapter.js
```diff
@@ -445,12 +516,17 @@ export function parse({minim, source, generateSourceMap}, done) {
         }
 
         // Currently, default responses are not supported in API Description format
-        // TODO: [Annotation] Add warning about not showing the default!
         const relevantResponses = _.chain(methodValue.responses)
           .omit('default')
           .omit(isExtension)
           .value();
 
+        if (methodValue.responses && methodValue.responses.default) {
```
smizell: Just a note, I am in the habit of this as well, but if you have lodash, you can do `_.get(methodValue, 'responses.default`. This saves you from the situation where `responses` is defined, but not an object.

I actually go an error today because of how we checked for some properties. I think we could resolve that with lodash.

Again, out of scope of this PR, but just mentioning.

test/fixtures/annotations.json
```diff
@@ -0,0 +1,430 @@
...
+    },
+    {
+      "element": "annotation",
+      "meta": {
+        "classes": [
+          "warning"
+        ]
+      },
+      "attributes": {
+        "code": 3,
```
smizell: Do we not distinguish between a warning an error in some way?

smizell:
@danielgtaylor Good to merge/publish. I added a few comments, but they are mostly out of scope (just wanted to think out loud, though).

README.md
danielgtaylor: I think this is a generally good idea, but definitely out of scope. Right now the codes are set up this way because of the refract spec, so we'll need to go back and modify that first.

src/adapter.js
danielgtaylor: Yeah, we could use a link, add a snippet describing the linked doc to the description/copy or we could just inline the content. I don't know which would be best (maybe an option makes sense?)

src/adapter.js
danielgtaylor: I like this idea, just not sure if it makes sense to add a new dependency for just this...

smizell: My mistake! Thought lodash was already there. Please ignore me in these situations :)

PR #21, opened 2015-12-02
Encode dash in URI templates

danielgtaylor:
This should fix an issue where `-` was mistakenly included as it is an invalid character in URI template variable names.

It is the responsibility of the tool displaying the variable names to decode them!

cc @smizell @netmilk

### DIFF
--- a/src/uri-template.js
+++ b/src/uri-template.js
@@ -16,7 +16,13 @@ export default function buildUriTemplate(basePath, href, pathObjectParameters = 
     const parameterNames = _.unique([].concat(pathObjectParameterNames, queryParameterNames));
     const parameterNamesString = parameterNames.length ? `{?${parameterNames.join(',')}}` : '';
 
-    return `${basePath}${href}${parameterNamesString}`;
+    const full = `${basePath}${href}${parameterNamesString}`;
+
+    // Before returning, we replace instances of `-` with `%2d`, but only when
+    // they occur inside of a template variable.
+    return full.replace(/\{.*?\}/g, (match) => {
+      return match.replace('-', '%2d');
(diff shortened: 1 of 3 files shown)


src/uri-template.js
```diff
@@ -16,7 +16,13 @@ export default function buildUriTemplate(basePath, href, pathObjectParameters =
     const parameterNames = _.unique([].concat(pathObjectParameterNames, queryParameterNames));
     const parameterNamesString = parameterNames.length ? `{?${parameterNames.join(',')}}` : '';
 
-    return `${basePath}${href}${parameterNamesString}`;
+    const full = `${basePath}${href}${parameterNamesString}`;
+
+    // Before returning, we replace instances of `-` with `%2d`, but only when
+    // they occur inside of a template variable.
+    return full.replace(/\{.*?\}/g, (match) => {
```
smizell: My question I had before was, should this be done for the full URI? Don't dashes only need to be replaced in the variables and not in the `basePath` and `href`?


In other words, in your tests, if you set the `basePath` to `/my-api`, it would convert that dash as well. Thoughts?

smizell:
Just one comment. If this is a non-issue, this is good to merge.

src/uri-template.js
danielgtaylor: I've updated the test to show that this works as expected. It only replaces things within `{` and `}` (non-greedy), which should only get variables. We can't just do the query vars because the path itself may contain variables, e.g. `pets/{id}{?foo}`.

smizell: I see now, thanks :)

smizell:
:+1:

PR #22, opened 2015-12-04
Add link relations to annotations

danielgtaylor:
This change adds link relations to annotations in the parser output. Right now it supports the `yaml-parser` and `refract-not-supported` relations.

cc @smizell

### DIFF
--- a/src/adapter.js
+++ b/src/adapter.js
@@ -11,14 +11,17 @@ const ANNOTATIONS = {
   CANNOT_PARSE: {
     type: 'error',
     code: 1,
+    relation: 'yaml-parser',
   },
   AST_UNAVAILABLE: {
     type: 'warning',
     code: 2,
+    relation: 'yaml-parser',
   },
   DATA_LOST: {
     type: 'warning',
     code: 3,
+    relation: 'refract-not-supported',
   },
 };
 
@@ -157,11 +160,17 @@ function makeSourceMap(SourceMap, ast, element, path) {
 }
 
 // Make a new annotation for the given path and message
-function makeAnnotation(Annotation, SourceMap, ast, result, info, path, message) {
+function makeAnnotation(Annotation, Link, SourceMap, ast, result, info, path, message) {
   const annotation = new Annotation(message);
   annotation.classes.push(info.type);
   annotation.code = info.code;
   result.content.push(annotation);
+
+  if (info.relation) {
+    const link = new Link();
+    link.relation = info.relation;
+    annotation.links.push(link);
+  }
 
   if (ast && path) {
     const position = getPosition(ast, path);
@@ -257,6 +266,7 @@ export function parse({minim, source, generateSourceMap}, done) {
   const Category = minim.getElementClass('category');
   const HrefVariables = minim.getElementClass('hrefVariables');
   const HttpHeaders = minim.getElementClass('httpHeaders');
+  const Link = minim.getElementClass('link');
   const MemberElement = minim.getElementClass('member');
   const ParseResult = minim.getElementClass('parseResult');
   const Resource = minim.getElementClass('resource');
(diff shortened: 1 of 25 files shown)


test/fixtures/annotations.json
```diff
@@ -231,6 +231,16 @@
       "meta": {
         "classes": [
           "warning"
+        ],
+        "links": [
+          {
+            "element": "link",
+            "meta": {},
+            "attributes": {
+              "relation": "refract-not-supported"
```
smizell: In using links, the plan was to use `origin` as the link relation and then a link to the origin of the issue.

```js
{
  element: 'link',
  attributes: {
    relation: 'origin',
    href: 'http://docs.apiary.io/parser/#refract-not-supported'
  }
}
```

Or something like that.

src/adapter.js
```diff
@@ -11,14 +11,22 @@ const ANNOTATIONS = {
   CANNOT_PARSE: {
     type: 'error',
     code: 1,
+    relation: 'yaml-parser',
   },
   AST_UNAVAILABLE: {
     type: 'warning',
     code: 2,
+    relation: 'yaml-parser',
```
smizell: I think I may have gotten confused before by this. Here, you call this a relation, but later you use it in the `href` of a link. Wondering if you could name it to something different to make it clearer.

src/adapter.js
```diff
@@ -157,15 +165,22 @@ function makeSourceMap(SourceMap, ast, element, path) {
...
   const annotation = new Annotation(message);
   annotation.classes.push(info.type);
   annotation.code = info.code;
   result.content.push(annotation);
 
+  if (info.relation) {
+    const link = new Link();
+    link.relation = 'origin';
+    link.href = `http://docs.apiary.io/validations/swagger#${info.relation}`;
+    annotation.links.push(link);
```
smizell: **Out of scope question—beware**

Could we improve the API here? 

```js
const link = annotations.links.add('link'); // automatically initializes and pushes to content
link.relation = 'origin';
link.href = `http://docs.apiary.io/validations/swagger#${info.relation}`;
```

We could then have a list of registered elements to add to content for arrays. Seems very common.

danielgtaylor: :+1:

src/adapter.js
danielgtaylor: Sure, I can update this. Sorry about that - it was set as a relation before.

src/adapter.js
```diff
@@ -157,15 +165,22 @@ function makeSourceMap(SourceMap, ast, element, path) {
...
+  }
+
   if (ast && path) {
     const position = getPosition(ast, path);
-    if (position) {
+    if (position && position.start !== undefined && !isNaN(position.start) && position.end !== undefined && !isNaN(position.end)) {
```
smizell: Do you need the `!== undefined` checks? Doesn't `!isNaN(undefined)` return false?

src/adapter.js
```diff
@@ -284,24 +301,60 @@ export function parse({minim, source, generateSourceMap}, done) {
...
 
-  $RefParser.dereference(loaded, (err, swagger) => {
+  // Some sane defaults since these are sometimes left out completely
+  if (loaded.info === undefined) {
+    loaded.info = {};
+  }
+
+  if (loaded.paths === undefined) {
+    loaded.paths = {};
+  }
```
smizell: Nice

src/adapter.js
```diff
@@ -284,24 +301,60 @@ export function parse({minim, source, generateSourceMap}, done) {
...
-      return done(err, parseResult);
+      if (swagger === undefined) {
+        return done(err, parseResult);
+      }
+
+      // Non-fatal errors, so let us try and create annotations for them and
+      // continue with the parsing as best we can.
+      if (err.details) {
+        const queue = [err.details];
+        while (queue.length) {
```
smizell: Curious: why a `while` block instead of `for` loop?

smizell:
@danielgtaylor This is not blocking at all, but if what I'm saying is correct, you may be able to change this:

https://github.com/apiaryio/fury-adapter-swagger/pull/22#discussion_r46875263

If not, I'll merge.

src/adapter.js
danielgtaylor: I went with a `while` because the length may change as we `push` / `shift` / `unshift` from the queue. It could be a `for` loop as long as the updated length is checked each iteration. I guess I don't have a strong opinion here, just seemed like the right thing to use.

src/adapter.js
danielgtaylor: I think you are right.

src/adapter.js
smizell: Wasn't really suggesting to change it, mostly curious when this is helpful :)

PR #24, opened 2015-12-08
Add support for x-summary and x-description

smizell:
Add support for x-summary and x-description on Path Item Objects.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Support `x-summary` and `x-description` in Path Item Objects
 - Generate Swagger schema and spec validation annotations.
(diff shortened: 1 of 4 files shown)


smizell:
cc @danielgtaylor

pksunkara:
@smizell Why not use `_summary` and `_description`? or maybe even 2 underscores? `x-summary` looks like it's related to HTTP.

kylef:
@pksunkara see http://swagger.io/specification/#vendorExtensions

smizell:
This allows us to add titles and descriptions in places Swagger does not
allow. We are using Swagger's vendor extension and their naming, per
Adam's. story.

PR #25, opened 2015-12-09
Do not encode JSON objects again

XVincentX:
This PR makes sure to check out what kind of parameter we're treating when creating a message body.
If it's an object, then we can serialise it as JSON, but if it's not, we should just forward it to the `Asset` object. If not, we will double encode it and a `JSON.parse` conversion will not give the desired result.


XVincentX:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;X9c4SgAK&#x2F;98-fix-json-examples"> Fix JSON examples</a>

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+; Install the "EditorConfig" plugin into your editor to use
+
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
```
smizell: A [Rider](https://en.wikipedia.org/wiki/Rider_(legislation)) :)

Looks good to me though.

src/adapter.js
```diff
@@ -695,7 +695,13 @@ export function parse({minim, source, generateSourceMap}, done) {
 
             // Responses can have bodies
             if (responseBody !== undefined) {
-              const bodyAsset = new Asset(JSON.stringify(responseBody, null, 2));
+              let formattedResponseBody = responseBody;
+
+              if (typeof(responseBody) === 'object') {
```
kylef: Would array's of objects also need to be stringified too?

smizell: Does this work? Can't valid JSON be any primitive value, from string, to array, to object? I will need to look at the Swagger spec, but I'm thinking we'll need to stringify anything other than a string (assuming a string is either a primitive string or JSON-stringified).

test/fixtures/refract/payload-as-string.json
```diff
@@ -0,0 +1,208 @@
...
+  "meta": {},
+  "attributes": {},
+  "content": [
+    {
+      "element": "category",
+      "meta": {
+        "classes": [
+          "api"
+        ],
+        "title": "Not formated example"
```
smizell: Nit: formatted misspelled

test/fixtures/refract/x-summary-and-description.json
```diff
@@ -52,7 +52,9 @@
     {
       "element": "category",
       "meta": {
-        "classes": ["api"]
+        "classes": [
+          "api"
+        ]
```
smizell: Specific reason for this?

pksunkara: It's how the generated JSONs are serialized.

smizell: True, but I'm not sure why this is a change to this PR as it seems unrelated. Is it making things consistent? Or was this making tests fail?

XVincentX: Well it was regenerated by `npm` test command. Will fix!

test/fixtures/refract/payload-as-string.json
XVincentX: Thanks, will fix

src/adapter.js
XVincentX: Right, I will revert the condition then-

test/fixtures/swagger/payload-as-string.yaml
```diff
@@ -0,0 +1,47 @@
...
+                "results": [
+                  {
+                    "id": 12312323,
+                    "type": "article"
+                  }
+                ]
+              }
+
+        default:
+          description: Error
```
XVincentX: True that, will fix.

test/fixtures/refract/x-summary-and-description.json
pksunkara: I don't think this needs to be fixed.

smizell: No problem to fix if it was auto generated. Was just curious why it was there.

smizell: I mean, you don't need to fix it.

XVincentX:
@smizell Thanks for the review. I've addressed all the comments; changed the condition, added new lines where needed and stripped down the fixture to the minimal possible one. Please have a new look, thanks!

PR #30, opened 2015-12-11
Add source maps for YAML parse errors

danielgtaylor:
There are two caveats:

1. We only get the *first* error.
2. We only get a single position, so the source map points to one character.

cc @smizell


src/adapter.js
```diff
@@ -332,6 +332,13 @@ export function parse({minim, source, generateSourceMap}, done) {
   } catch (err) {
     makeAnnotation(Annotation, Link, SourceMap, null, parseResult,
       ANNOTATIONS.CANNOT_PARSE, null, 'Problem loading the input');
+
+    if (err.mark) {
+      parseResult.content[0].attributes.set('sourceMap', [
```
smizell: You may want to use `.first()`?

danielgtaylor:
@smizell updated based on comments :smile:

PR #31, opened 2015-12-12
Refactor code

danielgtaylor:
See the changelog for details. All tests continue to pass :smile:

There's still a *lot* to do to make it more maintainable, but this I believe lays a good framework for doing so.

cc @smizell


src/ast.js
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      piece = pieces.shift();
+    }
+
+    return {start, end};
+  }
+}
+
+module.exports = Ast;
```
smizell: Could you use the ES2015 exports?

src/parser.js
```diff
@@ -0,0 +1,782 @@
...
+      this.makeAnnotation(ANNOTATIONS.CANNOT_PARSE, null,
+        'Problem loading the input');
+
+      if (err.mark) {
+        parseResult.first().attributes.set('sourceMap', [
+          new SourceMap([[err.mark.position, 1]]),
+        ]);
+      }
+
+      return done(null, parseResult);
```
smizell: Should you pass the `err` here?

src/parser.js
```diff
@@ -0,0 +1,782 @@
...
+    // Current resource group, if any
+    this.group = null;
+  }
+
+  parse(done) {
+    const {
+      Copy, Category, Member: MemberElement, ParseResult, SourceMap,
+    } = this.minim.elements;
+
+    const parser = new SwaggerParser();
```
smizell: This is now a little confusing in the context of a parser. You have a generic `Parser` class with a `parse` method that instantiates a `parser`. Is there anything that could be done to make clearer what is going on to the user?

src/parser.js
```diff
@@ -0,0 +1,782 @@
...
+      api.classes.push('api');
+
+      if (swagger.info) {
+        this.path.push('info');
+
+        if (swagger.info.title) {
+          api.meta.set('title', swagger.info.title);
+
+          if (this.generateSourceMap) {
+            this.makeSourceMap(api.meta.get('title'), this.path.concat(['title']));
```
smizell: This makes me feel that this `Parse` class is doing more than simply parsing, as in it has multiple responsibilities. Passing state between functions with `this` confuses me when I look at code :)

In other words, you could have a function that you pass a `Parse` instance into along with these other things and still have the same benefit without coupling the additional functionality to the instance.

This is mostly my opinion/preference. What are your thoughts?

src/parser.js
```diff
@@ -0,0 +1,782 @@
...
+      // The key is the href
+      _.each(_.omit(swagger.paths, isExtension), this.handlePath.bind(this));
+
+      done(null, parseResult);
+    });
+  }
+
+  // == Internal properties & functions ==
+
+  handlePath(pathValue, href) {
```
smizell: I was a tad confused of the naming of this. The `Parser` class has a property called `path`, but this is handling a different `paths` from Swagger.

smizell:
I think #32 requires a rebase, but it doesn't look too bad.

I think my only concern here is how much the parse class is doing. Of course people have different ways of doing classes and objects, so this may be preference. I personally try to have each class have a single responsibility. Additionally, the only real public interface I see here is `#parse`, and most of the others seem to be either basic functions or their own classes.

I'm really fine either way :) It's definitely a big improvement!

src/parser.js
danielgtaylor: Yeah, looks like it was fixed in the other PR. I'll rebase.

src/parser.js
danielgtaylor: I'm definitely open to ideas.

src/parser.js
danielgtaylor: Good catch. I'll update this name to make it more clear.

src/ast.js
danielgtaylor: Whoops, will update.

src/parser.js
smizell: Just wondering if you can call it something swagger-specific to not overload the term "parse" here.

PR #32, opened 2015-12-14
Return error on invalid YAML

JackuB:
This should fix false positive for invalid YAML files. 
It should also return error on invalid YAML.

### DIFF
--- a/test/adapter.js
+++ b/test/adapter.js
@@ -68,13 +68,6 @@ describe('Swagger 2.0 adapter', () => {
     });
   });
 
-  it('returns error for bad input', (done) => {
-    fury.parse({source: 'swagger: "2.0"\nbad: }'}, (err) => {
-      expect(err).to.exist;
-      done();
-    });
-  });
-
   context('can parse Swagger object', () => {
     const source = {swagger: '2.0', info: {title: 'Test'}};
     let result;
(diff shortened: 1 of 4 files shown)


JackuB:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;PUO1sCWR&#x2F;65-yaml-errors-and-warnings-showing-in-the-editor"> YAML errors and warnings showing in the editor</a>

test/adapter.js
```diff
@@ -104,6 +97,23 @@ describe('Swagger 2.0 adapter', () => {
     });
   });
 
+  context('cannot parse invalid Swagger YAML', () => {
+    const source = 'swagger: "2.0"\nbad: }';
+    let result;
+
+    before((done) => {
+      fury.parse({source}, (err) => {
+        result = err;
```
smizell: Is the correct naming of this variable? Wouldn't the result of the parse be called the `result`, but you are concerned with the `err` here?

smizell:
@JackuB Just one small naming question.

test/adapter.js
JackuB: I meant it as "a result of parse()", but let's give it a proper name.

JackuB:
@smizell updated

PR #35, opened 2015-12-15
Continue work to break up parser

danielgtaylor:
This does some more work to break up the Swagger parser into smaller methods and introduces the `withPath` method to make handling of YAML AST paths and source maps easier. It also adds documentation in a few places and a couple of `TODO` items we should look at (in another PR, probably).

cc @smizell

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -61,27 +61,27 @@ export default class Parser {
 
   parse(done) {
     const {
-      Copy, Category, Member: MemberElement, ParseResult, SourceMap,
+      Category, ParseResult, SourceMap,
     } = this.minim.elements;
-
     const swaggerParser = new SwaggerParser();
-    const parseResult = new ParseResult();
-    this.result = parseResult;
 
+    this.result = new ParseResult();
+
+    // First, we load the YAML if it is a string, and handle any errors.
     let loaded;
     try {
       loaded = _.isString(this.source) ? yaml.safeLoad(this.source) : this.source;
     } catch (err) {
-      this.makeAnnotation(ANNOTATIONS.CANNOT_PARSE, null,
+      this.createAnnotation(ANNOTATIONS.CANNOT_PARSE, null,
         (err.reason || 'Problem loading the input'));
 
       if (err.mark) {
-        parseResult.first().attributes.set('sourceMap', [
+        this.result.first().attributes.set('sourceMap', [
           new SourceMap([[err.mark.position, 1]]),
         ]);
       }
 
-      return done(new Error(err.message), parseResult);
+      return done(new Error(err.message), this.result);
     }
 
     // Some sane defaults since these are sometimes left out completely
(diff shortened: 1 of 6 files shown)


src/parser.js
```diff
@@ -125,188 +126,237 @@ export default class Parser {
...
+        this.createSourceMap(i, this.path);
+      }
     }
 
-    // Provide users a way to add a description to a resource in Swagger
-    if (pathValue['x-description']) {
-      const resourceDescription = new Copy(pathValue['x-description']);
-      resource.push(resourceDescription);
+    for (i = 0; i < args.length - 1; i++) {
+      this.path.pop();
```
smizell: Just a question: would slice be better than a loop and multiple pops? I personally have no idea, but maybe you know.

src/parser.js
```diff
@@ -125,188 +126,237 @@ export default class Parser {
...
     }
+  }
 
-    if (this.useResourceGroups()) {
-      const groupName = pathValue['x-group-name'];
+  // This is like `withPath` above, but slices the path before calling by
+  // using the first argument as a length (starting at index 0).
+  withSlicedPath(...args) {
```
smizell: lol ^ see my comment above

src/parser.js
```diff
@@ -401,86 +425,118 @@ export default class Parser {
...
+    });
+  }
 
-        if (responseValue.examples) {
-          examples = responseValue.examples;
-        }
+  // Convert a Swagger response & status code into Refract transactions.
+  handleSwaggerResponse(resource, transition, method, transitionParameters, responseValue, statusCode) {
+    let examples = {
+      '': undefined,
```
smizell: What does this do?

smizell:
@danielgtaylor I just had a couple of questions, then this is good to go.

Also a side though, I'm wondering if `async.each` would make things faster here. We do a lot of `_.each` and `.forEach` that I'm betting we could speed things up. But that's another dependency.

src/parser.js
danielgtaylor: Originally my goal was to keep the reference to the array. E.g. if you do `let foo = this.path` and then push/pop on the path then `foo` still points to the same instance. If we slice, we need to replace it with the new (sliced) copy so `foo` is now not the same as `this.path` anymore. I don't know if that matters, but it seemed less surprising to me using push/pop.

src/parser.js
danielgtaylor: Heh, the behavior here is different though - it slices *before* so that you can remove items from the current path. It's only used in one place right now. After running, the `this.path` is the same instance as when it started, so at completion the behavior is consistent with the method above. E.g.

```js
let foo = this.path;

this.withPath('bar', () => {});
assert.equal(foo, this.path); // True

this.withSlicedPath(1, 'baz', () => {});
assert.equal(foo, this.path); // True
```

src/parser.js
danielgtaylor: I'll add a comment to the source, but the gist is that the Refract spec *requires* an HTTP Transaction to have a request and a response, so this seeds the examples list to create a blank request/response. Why not just leave out the transaction, you ask? That is the only way we have of specifying which HTTP methods are allowed, so we need to create a transaction.

danielgtaylor:
@smizell thanks for the review! Regarding `async`, it won't actually speed things up here. Nothing we are doing is IO-bound or asynchronous (after validation). We could certainly use `async.parallel` (maybe as a code organization or flow control utility) but in reality it won't run any of it in parallel, just one after the other. It only works in parallel if the code path can return, e.g. if your functions call one of the `fs` or `http` or some database module methods that register a callback.

src/parser.js
smizell: Gotcha, makes sense.

This is something I could see being wrapped in some class to hide state yet be a single reference to this functionality.

PR #36, opened 2016-01-06
Prevent bug overwriting resource href template.

danielgtaylor:
This change prevents an issue where multiple transitions on a resource could
lead to the resource `href` being overwritten as each transition is processed.
The error occurs when transitions define href variables (params in Swagger).

I believe the correct behavior is to set an `href` on the resource based on
the resource href variables, then for each transition if it has a different
`href` it will get set on the transition.

**Note**: downstream tooling probably needs to be updated to make use of
the transition's `href` attribute.

cc @smizell @Baggz

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -154,6 +154,11 @@ export default class Parser {
 
   // == Internal properties & functions ==
 
+  // Base path (URL) name for the API
+  get basePath() {
+    return (this.swagger.basePath || '').replace(/[/]+$/, '');
+  }
+
   // Lazy-loaded input AST is made available when we need it. If it can't be
   // loaded, then an annotation is generated with more information about why.
   get ast() {
@@ -315,6 +320,13 @@ export default class Parser {
         resource.hrefVariables = resourceHrefVariables;
       }
 
+      // Set the resource-wide URI template, which can further be overridden
+      // by individual transition URI templates. When creating a transition
+      // below, we *only* set the transition URI template if it differs from
+      // the one we've generated here.
+      const hrefForResource = buildUriTemplate(this.basePath, href, pathObjectParameters);
+      resource.attributes.set('href', hrefForResource);
+
       // TODO: It should add support for `body` and `formData` parameters as well.
       if (pathObjectParameters.length > 0) {
         pathObjectParameters.forEach((parameter, index) => {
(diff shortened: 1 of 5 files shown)


src/parser.js
```diff
@@ -154,6 +154,11 @@ export default class Parser {
 
   // == Internal properties & functions ==
 
+  // Base path (URL) name for the API
+  get basePath() {
+    return (this.swagger.basePath || '').replace(/[/]+$/, '');
```
smizell: Is this the correct move? In other words, can we safely assume the user will always put their slashes on the front of the  `href` and not on the end of `basePath`?

- Base Path: `http://example.com/`
- Resource HREF: `users/{id}`

src/parser.js
```diff
@@ -365,13 +377,13 @@ export default class Parser {
...
-      const hrefForResource = buildUriTemplate(basePath, href, resourceParameters, queryParameters);
-      // TODO: This is probably a bug. Transitions can have their own `href` and
-      // we should use that instead of resetting the resource's href each time.
-      // The open question is do resources still get an href? Then we should
-      // only set it here on the transition if it is different.
-      resource.attributes.set('href', hrefForResource);
+      // Here we generate a URI template specific to this transition. If it
+      // is different from the resource URI template, then we set the
+      // transition's `href` attribute.
+      const hrefForTransition = buildUriTemplate(this.basePath, href, resourceParameters, queryParameters);
```
smizell: Does this add any parameters from the resource to this? In other words, a resource can have X query parameters and a transition can have X + Y query parameters.

test/fixtures/refract/path-level-params.json
```diff
@@ -115,13 +115,14 @@
...
             },
-            "href": "/test/{id}{?search,arg}"
+            "href": "/test/{id}{?search}"
           },
           "content": [
             {
               "element": "transition",
               "meta": {},
               "attributes": {
+                "href": "/test/{id}{?search,arg}",
```
smizell: Looks like you cover the situation here I mentioned above, where a transition adds to the query parameters.

src/parser.js
danielgtaylor: The above behavior is correct according to the spec as far as I can tell:

> The field name MUST begin with a slash. The path is appended to the basePath in order to construct the full URL.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#paths-object

smizell: You're right here then.

smizell:
@danielgtaylor This looks right to me. I just want to double check, because it looks before we were not adding certain `hrefVariables` to the `resource` meta values, but is it now correct in how it displays `hrefVariables`, both for resources and transitions?

If so, I think you fixed this. Now my brain hurts again, though.

danielgtaylor:
Merged after discussion with @smizell and @Baggz

PR #37, opened 2016-01-12
Implement source map information for bad references

danielgtaylor:
Previous behavior:

* Internal error is raised, missing `err.details` so no annotations are generated.
* Processing continues as if nothing happened (swallowed error).
* Output is weird since it includes unresolved `$ref` keys

After this PR:

* Internal error is raised, missing `err.details` so we generate an annotation based on the error itself. If a regex can match the error message to get a reference name for a bad reference, then it creates a source map for the annotation.
* The error and result-with-annotation are returned instead of continuing to process. The user needs to fix the bad reference first.
* Output is minimal - really just the error.

cc @smizell @Baggz

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+- Bad JSON references (using `$ref`) are now returned as parse result
+  annotations that include source maps which point to the `$ref` line, making
+  debugging easier.
+
 # 0.6.1 - 2016-01-11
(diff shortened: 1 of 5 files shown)


pksunkara:
:+1:

smizell:
Looks good to me :)

danielgtaylor:
Published in version 0.6.2

PR #38, opened 2016-01-18
Add an example for form-data parameters

danielgtaylor:
No code changes, just a new example showing a Swagger file with path-level `formData` parameters and how they could be appended/overwritten by request-level `formData` parameters. This helped me to debug an issue and confirm that the problem lies outside of Fury. I figure it'll be a good example to keep in the set.

@Baggz @smizell


kylef:
:+1:

danielgtaylor:
Merged, but no release required since it's only a test case.

PR #39, opened 2016-01-26
Treat JSON Schema validation errors as errors

kylef:
If the Swagger document didn't validate against the JSON Schema, we previously continued to translate the Swagger document to Refract.

This change stops that behaviour and instead stops the translation. The problems with translating the "invalid" document is that many types in the Swagger document might not be correct and cause our translation to fail and throw JS exceptions. For example if a parameter was not an object but instead none, or an array.

For example, before we would fail due to mis-matched types. Here's an example of what would fail before:

```yaml
swagger: '2.0'
info:
  title: Bad Parameters
  version: '1.0'
paths:
  '/':
    get:
      responses:
        200:
          examples:
            foo: t
    parameters:
      -
```

```
/app/node_modules/fury-adapter-swagger/lib/parser.js:478 
           return parameter['in'] === 'query'; 
                           ^ 
 TypeError: Cannot read property 'in' of null 
     at /app/node_modules/fury-adapter-swagger/lib/parser.js:478:27 
```


test/fixtures/swagger/param-no-response.yaml
```diff
@@ -1,4 +1,7 @@
 swagger: '2.0'
+info:
+  title: Parameters with no response
```
kylef: Having no responses would result in this being a invalid Swagger document. @danielgtaylor does it make sense to still have this fixture given it results in an invalid Swagger document?

Given the changes in this PR, it seems unnecessary, but wanted to check before removing.

danielgtaylor: In this case I think it can be removed! :+1:

danielgtaylor:
I think generally this looks great. Thanks for the change! Just a couple high-level comments:

1. Can you also update the README as annotation code `4` is now an error, not a warning?
2. I'm a little worried about the coverage drop. Is there any way we can either create a test case to get the coverage up again or remove some of the existing inline error handling code that is no longer needed with the new behavior?

Aside from those two, this looks ready to go :cake:

pksunkara:
I don't know why the coverage decreased by so much. Is coveralls having any kind of issue?

Issue #45, opened 2016-04-04
Add recursive JSON Schema support


pksunkara:
The following swagger document is crashing the module.

``` yaml
swagger: '2.0'
info:
  description: Sample API
  version: 1.0.0
  title: Sample API
host: 'example.com'
basePath: /v1
tags:
  - name: Recursion
    description: Lorem Ipsum
schemes:
  - https
paths:
  '/recursive':
    get:
      tags:
        - Recursion
      summary: Recursion test
      description: |-
        This is a test to show that there are issues with recursion
      parameters:
        - name: tag
          in: query
          description: 'Test 1 2'
          required: false
          type: string
      responses:
        '200':
          description: Success
          schema:
            "$ref": "#/definitions/category"
definitions:
  category:
    type: object
    properties:
      id:
        type: string
      categoryAsArray:
        type: array
        description: A list of categories
        items:
          "$ref": "#/definitions/category"
      categoryAsObject:
        type: object
        "$ref": "#/definitions/category"
    description: A category groups one or more categories
```

kylef:
To add more information:

```
/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:1008
      var schemaAsset = new Asset(JSON.stringify(jsonSchema));
                                       ^
TypeError: Converting circular structure to JSON
    at Object.stringify (native)
    at Parser.createAssetFromJsonSchema (/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:1008:40)
    at Parser.<anonymous> (/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:697:40)
    at Parser.withPath (/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:251:54)
    at Parser.withSlicedPath (/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:303:21)
    at Parser.<anonymous> (/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:696:35)
    at Parser.withPath (/usr/local/lib/node_modules/swagger2blueprint/node_modules/fury-adapter-swagger/lib/parser.js:251:54)
[...]

PR #46, opened 2016-04-12
Catch all JS exceptions at the beginning

pksunkara:
Fixes the crash in #45

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -30,6 +30,10 @@ const ANNOTATIONS = {
     code: 4,
     fragment: 'swagger-validation',
   },
+  UNCAUGHT_ERROR: {
+    type: 'error',
+    code: 5,
+  },
 };
 
 // Test whether a key is a special Swagger extension.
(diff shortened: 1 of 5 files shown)


src/parser.js
```diff
@@ -30,6 +30,10 @@ const ANNOTATIONS = {
     code: 4,
     fragment: 'swagger-validation',
   },
+  UNCAUGHT_ERROR: {
```
kylef: Remember to update the table in the READMe including the new code https://github.com/apiaryio/fury-adapter-swagger#parser-codes

src/parser.js
```diff
@@ -148,29 +152,36 @@ export default class Parser {
...
+        // Swagger has a paths object to loop through that describes resources
+        _.each(_.omit(swagger.paths, isExtension), (pathValue, href) => {
+          this.handleSwaggerPath(pathValue, href);
+        });
 
-      done(null, this.result);
+        return done(null, this.result);
+      } catch (exception) {
+        this.createAnnotation(ANNOTATIONS.UNCAUGHT_ERROR, null,
+          ('Problem converting the input into refract due to swagger support still being in beta'));
```
kylef: I don't think this message is very user friendly.

What about:

> There was a problem converting the Swagger document, the document may be using unsupported features in the Swagger Beta.

pksunkara:
Updated.

src/parser.js
```diff
@@ -148,29 +152,36 @@ export default class Parser {
...
+          this.handleSwaggerPath(pathValue, href);
+        });
 
-      done(null, this.result);
+        return done(null, this.result);
+      } catch (exception) {
+        this.createAnnotation(ANNOTATIONS.UNCAUGHT_ERROR, null,
+          ('There was a problem converting the Swagger document, the document may be using unsupported features in the Swagger Beta'));
+
+        return done(new Error(exception.message), this.result);
```
kylef: Should we return the result as it is? Perhaps it will have broken / unknown state due to the exception?

pksunkara: It will be in a good state. It will have stuff until the stuff got added and that's all. You can check the test case.

Baggz:
@pksunkara @kylef 

I'll leave the final review for @danielgtaylor 

But as far as I see this is safe—it won't break Editor Kit (used to display semantic issues and beta warnings in the API Editor), nor Apiary.

danielgtaylor:
LGTM guys :+1:

PR #48, opened 2016-04-25
Use swagger-zoo

pksunkara:


test/adapter.js
```diff
@@ -32,7 +29,12 @@ function testFixture(description, filename, subDir = true, generateSourceMap = f
       // Invoke with the env var GENERATE set to regenerate the fixtures.
       if (process.env.GENERATE) {
         expected = output.toRefract();
-        fs.writeFileSync(path.join(__dirname, expectedName), JSON.stringify(expected, null, 2), 'utf8');
+
+        if (generateSourceMap) {
```
danielgtaylor: I'm not sure this makes sense. Maybe we should just remove the `GENERATE` option now? Really it should belong in the zoo repo instead.

pksunkara: @danielgtaylor That was difficult to do in swagger-zoo. I think we should leave it here.

danielgtaylor: Spoke on Slack, and the point I am missing is that there are setters that *do* save the files in the zoo. This makes sense then, so let's keep it as-is.

PR #52, opened 2016-05-13
Use DATA_LOST error for circular JSON Schema Refs

pksunkara:
Also add a fragment to uncaught_error

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.8.0-pre.0",
+  "version": "0.8.0-pre.2",
   "description": "Swagger 2.0 parser for Fury.js",
   "main": "./lib/adapter.js",
   "tonicExampleFilename": "tonic-example.js",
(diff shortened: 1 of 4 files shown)


smizell:
Looks good. Waiting for tests to pass.

smizell:
@pksunkara getting failing tests that look like they are related to the linter?

pksunkara:
@smizell Updated.

smizell:
@pksunkara I see coverage decreased 3.1%. That seems like a lot. Is the reason because the catch is not tested? How hard would it be to test it?

pksunkara:
@smizell Coverage decreased because I removed a few tests since they work with `auth` only,

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.8.0-pre.0",
+  "version": "0.8.0-pre.2",
```
kylef: Since you've changed the version number, should CHANGELOG be updated to reflect this too?

pksunkara: No, these are special pre-releases.

PR #56, opened 2016-05-31
Generate message bodies using schema for dredd to work with

pksunkara:
Depends on apiaryio/swagger-zoo#22 and needs publishing of swagger-zoo@2.1.0 and rebuilding the ci.

### DIFF
--- a/src/generator.js
+++ b/src/generator.js
@@ -2,7 +2,7 @@ import faker from 'json-schema-faker';
 import annotations from './annotations';
 
 faker.option({
-  useDefaultValue: true
+  useDefaultValue: true,
 });
 
 export function bodyFromSchema(schema, payload, parser) {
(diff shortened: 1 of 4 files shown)


src/generator.js
```diff
@@ -0,0 +1,29 @@
...
+
+    asset = new Asset(body);
+
+    asset.classes.push('messageBody');
+    asset.contentType = 'application/json';
+
+    payload.content.push(asset);
+  } catch (exception) {
+    parser.createAnnotation(annotations.DATA_LOST, parser.path,
+      'Unable to generate fake body out of schema');
```
kylef: I think this message could be a little more user-friendly. Not sure that everyone will understand what `fake body` means. Perhaps this could read:

> Unable to generate a JSON example message body from JSON Schema.

kylef:
One minor comment otherwise 👍

PR #58, opened 2016-06-03
Added produces/consumes headers and inferred links

pksunkara:


pksunkara:
Depends on release on https://github.com/apiaryio/swagger-zoo/pull/23

src/link.js
```diff
@@ -0,0 +1,25 @@
+export function baseLink(element, parser, options) {
+  const {Link} = parser.minim.elements;
+  options = options || {}
+
+  const link = new Link();
+  link.relation = options.relation || 'origin';
+  link.href = `http://docs.apiary.io/validations/swagger#${(options.fragment || '')}`;
```
kylef: These links are 404, should there be anything here? Shouldn't this be a link to a specification about this element type?

They are also against docs.apiary.io which was moved to help.apiary.io

pksunkara: We don't need to point them to anything, these are just links. That is all.

Also moving to help.apiary.io means @Baggz need to edit the editor too. So, I have left if for some later date.

kylef:
There are a bunch of failures in CI, mostly listing -> https://travis-ci.org/apiaryio/fury-adapter-swagger/builds/135036216

pksunkara:
@kylef Updated.

PR #59, opened 2016-06-07
Params with enum and default

pksunkara:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.8.0-pre.10",
+  "version": "0.8.0-pre.12",
```
honzajavorek: why jumping 2 versions in one PR?

src/parser.js
```diff
@@ -798,25 +798,83 @@ export default class Parser {
...
+
+    if (parameter.enum) {
       element = new ArrayElement();
+      element.element = 'enum';
 
-      if (parameter.items) {
-        element.content = [this.convertParameterToElement(
-          parameter.items, (path || []).concat(['items']), true)];
-      }
+      _.each(parameter.enum, (value, index) => {
```
honzajavorek: if `parameter.enum` is array, then you can use built-in `parameter.enum.forEach((value, index) => {`

pksunkara: I know. But we consistently use `_.each` in this repo.

package.json
pksunkara: Even is for master branch prereleases and Odd if for beta branch prereleases.

PR #62, opened 2016-06-08
Stringify generated message bodies. Fixes #61

### LINKED ISSUE #61: Message body generator produces invalid API Elements output
As reported in this [Trello card](https://trello.com/c/0vBs8g7R/114-fury-swagger-adapter-or-metamorphoses-doesn-t-stringify-response-bodies), sometimes an object will be given as a request/response body instead of a string, which can screw with the documentation, mock server, etc. I believe I have traced this back to the Swagger adapter. First, here is a sample minimal valid input document:

```yaml
swagger: "2.0"
info:
  version: 1.0.0
  title: Body Test
  description: A test to see how bodies get parsed / created
host: example.com
schemes:
  - http
consumes:
  - application/json
produces:
  - application/json
paths:
  /resource:
    get:
      description: Get a resource
      operationId: getResource
      responses:
        200:
          description: response description
          schema:
            example:
              id: 123
              name: Resource 1
```

Note that no response example is given directly. Instead, an example is set in the JSON Schema. Using the [Fury Swagger Adapter 0.8.0-pre.12 in Tonic](https://tonicdev.com/danielgtaylor/fury-adapter-swagger-bad-message-body) you get the following output excerpt:

```json
{
  "element": "asset",
  "meta": {
    "classes": [
      "messageBody"
    ],
    "links": [
      {
        "element": "link",
        "meta": {},
        "attributes": {
          "relation": "inferred",
          "href": "http://docs.apiary.io/validations/swagger#message-body-generation"
        },
        "content": []
      }
    ]
  },
  "attributes": {
    "contentType": "application/json"
  },
  "content": {
    "id": 123,
    "name": "Resource 1"
  }
},
```

This (inferred message body generation from JSON Schema) is apparently a new (and neat) feature. The problem I am having is that it is not [spec-compliant with API Elements and wreaks havoc downstream. The problematic part is the `content` of the asset, which is an `object` instead of `string`.

API Elements defines an `Asset` as having a `content` that [**must** be a string](http://api-elements.readthedocs.io/en/latest/element-definitions/#asset-base-api-element). The code should probably be updated to stringify the body similar to [this existing code](https://github.com/apiaryio/fury-adapter-swagger/blob/master/src/parser.js#L620-L622).

danielgtaylor:
This should fix the issue in #61, and can be tested with the associated change in apiaryio/swagger-zoo#25.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.8.0-pre.12",
+  "version": "0.8.0-pre.14",
   "description": "Swagger 2.0 parser for Fury.js",
   "main": "./lib/adapter.js",
   "tonicExampleFilename": "tonic-example.js",
@@ -30,7 +30,7 @@
     "coveralls": "^2.11.2",
     "fury": "^2.x.x",
     "peasant": "^0.5.2",
-    "swagger-zoo": "^2.1.2"
+    "swagger-zoo": "^2.1.3"
   },
   "author": "Apiary.io <support@apiary.io>",
   "license": "MIT"
--- a/src/generator.js
+++ b/src/generator.js
@@ -11,7 +11,11 @@ export function bodyFromSchema(schema, payload, parser) {
   let asset = null;
 
   try {
-    const body = schema.example || JSON.stringify(faker(schema), null, 2);
+    let body = schema.example || faker(schema);
+
+    if (typeof(body) !== 'string') {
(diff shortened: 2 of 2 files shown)


pksunkara:
Looks good but swagger-zoo@2.1.3 needs to be released before this is merged. And this PR should update package.json reflecting the above.

danielgtaylor:
Thanks! :+1:

PR #73, opened 2016-07-01
Authentication and 0.8.0

pksunkara:


src/parser.js
```diff
@@ -293,14 +296,231 @@ export default class Parser {
...
+              this.withPath('scopes', () => {
+                this.oauthScopes(element, item.scopes);
+              });
+            }
+
+            this.oauthTransitions(element, item);
+            break;
+
+          default:
+            break;
```
kylef: I think we should if we're given security definitions we don't understand. Informing the user that this scheme is not supported?

pksunkara: That is taken care by underlying swagger-parser

pksunkara:
This was already reviewed once. This is just a PR merge to master.

PR #74, opened 2016-07-13
Parse each Swagger path on separately in event loop

kylef:
We noticed that some code paths inside `handleSwaggerPath` perform slowly and we want to allow other events in the event loop to run sooner. We will now call `handleSwaggerPath` on separate event loop ticks which allows other events in the event loop to run before the complete parse is complete.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,11 @@
+# Master
+
+## Enhancements
+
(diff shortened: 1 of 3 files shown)


kylef:
Coverage unfortunately goes down because there is a code path for browsers which isn't tested in CI/coveralls.

w-vi:
:+1:

PR #75, opened 2016-07-26
Ability to add example values to parameters ('x-example' property)

honzajavorek:
This PR adds support for parameter property called `x-example`. The intended behavior is to:

- Add example value for  `header`, `formData`, `path`, and `query` parameters.
- Ignore `x-example` within `body` parameters and warn that `schema.example` should be used instead.
- The example value should get priority over the `default` value.
- The example value should get priority over the first of `enum` values. (In that case, it should be one of the specified values though).

---------------------

Tests are available in https://github.com/apiaryio/swagger-zoo/pull/29.
This PR is a part of https://github.com/apiaryio/dredd/issues/561 implementation.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,11 @@
+
+# 0.9.0 - 2016-07-27
+
+## Enhancements
+
+- Added support for parameter property called x-example, which allows to specify
+  example values also for non-body parameters.
+
 # 0.8.1 - 2016-07-16
 
 ## Enhancements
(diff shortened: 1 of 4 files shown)


kylef:
@honzajavorek Please add a changelog entry in the `CHANGELOG.md` file for this change and release.

honzajavorek:
@kylef This PR doesn't bump version. If changelog entry should be added, then I guess I should add also the version bump, correct?

pksunkara:
Yes, please add changelog and bump version.

pksunkara:
Please squash the last 2 commits.

PR #79, opened 2016-08-25
Upgrade to lodash 4

kylef:
Closes #78


kylef:
Please do not merge, I think dropping Node 0.10 support will be a problem for the core application.

PR #81, opened 2016-08-29
Exclude extensions from operations

kylef:
This PR fixes a bug where a certain Swagger document (see text fixture) can cause a crash in the adapter because the operation is not an object. Which eventually results in the following exception:

```
Aug 24 21:06:25:  2016-08-24T19:06:22.825Z	d895d769-6a2d-11e6-a48f-5953de4caa26	TypeError: Cannot read property 'tags' of null 
Aug 24 21:06:25:          at /var/task/node_modules/fury-adapter-swagger/lib/parser.js:1177:28 
Aug 24 21:06:25:          at /var/task/node_modules/fury-adapter-swagger/node_modules/lodash/index.js:3073:15
```

I've added the first regression test to the Swagger adapter and I don't think that regression tests belong in the Swagger Zoo so this adds code to run regression tests found in the test fixtures directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "chai": "^3.2.0",
+    "glob": "^6.0.1",
     "coveralls": "^2.11.2",
     "fury": "^2.x.x",
     "peasant": "^0.5.2",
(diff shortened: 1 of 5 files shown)


pksunkara:
@kylef Travis failing

kylef:
Patched.

pksunkara:
@kylef Please tag and release.

PR #85, opened 2016-09-01
fix: Handle empty elements in array while determining source map

kylef:
You can see that the test fixture has an empty array element, the previous code would error because it would try to grab the source map information for the empty element.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -62,8 +62,17 @@ export default class Ast {
 
         if (!pieces.length) {
           // This is the last item!
-          start = newNode.start_mark.pointer;
-          end = newNode.end_mark.pointer;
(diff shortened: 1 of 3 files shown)


pksunkara:
Let me see if I can improve upon this.

pksunkara:
This PR would fail if you add `- username` after `- id` in the given example. This is not exactly fixing the correct issue. The actual bug is in https://github.com/apiaryio/fury-adapter-swagger/blob/master/src/ast.js#L42. Apparently, the path is going into the `User` instead of `required, 1`

pksunkara:
Therefore, closing this PR. I will be using the test case from here though.

PR #87, opened 2016-09-20
Correct the sourcemap paths in convertParameterToMember

pksunkara:


package.json
```diff
@@ -31,7 +31,7 @@
     "coveralls": "^2.11.2",
     "fury": "^2.x.x",
     "peasant": "^0.5.2",
-    "swagger-zoo": "2.2.1"
+    "swagger-zoo": "2.3.x"
```
kylef: I'd keep this to an exact version. Any changes in the fixtures would break the tests from passing so it's best to pin to the exact version here.

src/parser.js
```diff
@@ -734,8 +736,29 @@ export default class Parser {
         transition.id = methodValue.operationId;
       }
 
+      /* eslint-disable prefer-const */
+      let resourceParametersWithPath = [];
```
kylef: Why can't this be const?

src/parser.js
```diff
@@ -734,8 +736,29 @@ export default class Parser {
...
+      /* eslint-enable prefer-const */
+
+      _.forEach(resourceParameters, (resourceParameter, index) => {
+        let inTransition = false;
+        const len = transitionParameters.length;
+
+        for (let i = 0; i < len; i++) {
```
kylef: Can you use `for of` syntax? much clearer than iterating via indexes.

kylef: Actually, I think it would be simpler to use a [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) for the matching parameter. There would be less state to track. 

Something like:

```js
const hasTransitionParameter = transitionParamters.some((parameter) => parameter.name == resourceParameter.name...);

if (!hasTransitionParameter) {
    // push param
} 
```

src/parser.js
```diff
@@ -734,8 +736,29 @@ export default class Parser {
         transition.id = methodValue.operationId;
       }
 
+      /* eslint-disable prefer-const */
+      let resourceParametersWithPath = [];
+      /* eslint-enable prefer-const */
+
+      _.forEach(resourceParameters, (resourceParameter, index) => {
```
kylef: Any reason not to use a standard for of loop in ES6 instead of using lodash loop?

src/parser.js
```diff
@@ -1310,21 +1331,22 @@ export default class Parser {
 
   // Create a new HrefVariables element from a parameter list. Returns either
   // the new HrefVariables element or `undefined`.
-  createHrefVariables(params) {
+  createHrefVariables(params, extraParamsWithPath) {
```
kylef: Please update the method documentation above to explain this new argument.

src/parser.js
pksunkara: Because `_.forEach` takes care of objects too unlike `for of` which can be finicky. So, we have standardised using `_.forEach`.

pksunkara:
Updated.

pksunkara:
> A Swagger operation which doesn't have parameters where the path has some parameters

There is a test for this already in `x-example` feature.

src/parser.js
```diff
@@ -734,10 +736,25 @@ export default class Parser {
         transition.id = methodValue.operationId;
       }
 
-      // For each uriParameter, create an hrefVariable
-      const methodHrefVariables = this.createHrefVariables(transitionParameters);
-      if (methodHrefVariables) {
-        transition.hrefVariables = methodHrefVariables;
+      if (transitionParameters.filter((p) => p.in === 'query' || p.in === 'path').length) {
```
pksunkara: Use `some`

pksunkara:
Unfortunately, I can't test the code in this PR using swagger-zoo because default value for path is `this.path` everywhere the function is being used.

I found the bug because I wanted to send a different path when merging the parameters and it didn't work.

pksunkara:
@kylef Added functional tests. This is now ready.

pksunkara:
Never mind about the codacy thing. I was just testing something.

kylef:
:+1:

PR #88, opened 2016-10-19
Limit array maxItems and string maxLength in generation from JSON Schema

kylef:
This adds limits to the value of `minItems`, `maxItems` for a JSON Schema array along with a limit of the `minLength` and `maxLength` of a string inside a JSON schema.

Otherwise the following JSON Schema will generate a huge output and take a considerable amount of time:

``` json
{
    "type": "array",
    "items": {
         "type": "string",
         "minLength": 100000,
         "maxLength": 10000000
    },
    "minItems": 1000,
    "maxItems": 10000000000
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "dependencies": {
     "babel-runtime": "^5.8.20",
     "js-yaml": "^3.4.2",
-    "json-schema-faker": "^0.3.3",
+    "json-schema-faker": "^0.3.7",
     "lodash": "^4.15.0",
     "swagger-parser": "^3.3.0",
     "yaml-js": "^0.1.3"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "babel-runtime": "^5.8.20",
     "js-yaml": "^3.4.2",
-    "json-schema-faker": "^0.3.3",
+    "json-schema-faker": "git+https://github.com/kylef/json-schema-faker.git#d64247ff",
```
kylef: Waiting on upstream PR.

pksunkara:
Is this a new bug we found?

kylef:
@pksunkara it's the only problem with the Swagger document I've been investigating in our current sprint. There appears to have been another problem that has been resolved in the parsing service, perhaps another update or an update to a dependency has resolved it.

PR #92, opened 2016-11-24
fix: Skip circluar references in examples

w-vi:


pksunkara:
I am not sure if the fixture is actually getting tested. Also, can you please make the release commit in this PR too?

w-vi:
The test gets tested : https://travis-ci.org/apiaryio/fury-adapter-swagger/builds/178621426#L534

w-vi:
Comments addressed.

src/parser.js
```diff
@@ -994,18 +994,22 @@ export default class Parser {
...
-              this.createSourceMap(bodyAsset, this.path);
-            }
+              if (this.generateSourceMap) {
+                this.createSourceMap(bodyAsset, this.path);
+              }
 
-            response.content.push(bodyAsset);
+              response.content.push(bodyAsset);
+            } catch (exception) {
+              this.createAnnotation(annotations.DATA_LOST,  this.path, 'Circular references in examples are not yet supported.');
```
kylef: Wouldn't this catch other exceptions and falsy show a warning about circular references?

pksunkara: Yeah, the `createSourceMap` might become an issue. Can you try catch only the JSON stringily please?

kylef: Or perhaps we can filter for the specific exception we want to catch and re-raise the rest.

w-vi:
Better?

PR #98, opened 2017-01-24
Introduce parameter inheritance

klokane:


test/inherit-parameters.js
```diff
@@ -0,0 +1,234 @@
...
+      });
+    });
+  });
+
+  context('Header Parameter', () => {
+    it('on Path', (done) => {
+      const source = makeSource('/');
+      source.paths['/'].parameters.push(makeParameter('test', 'header'));
+
+      doParse(source, done, (result) => {
```
klokane: @pksunkara as we spoke earlier, for this case is not created sourceMap, please, look at it

src/parser.js
```diff
@@ -265,16 +265,19 @@ export default class Parser {
   // a function (e.g. to create an element).
   withPath(...args) {
     let i;
+    const originalPath = _.clone(this.path);
```
klokane: Agree, but there is more parts which should be refactorized.
especialy `parser.handleSwaggerExampleRequest()`is good adept.

src/parser.js
```diff
@@ -889,13 +898,36 @@ export default class Parser {
         });
       });
 
+      _.forEach(resourceBodyParameters, (param) => {
+        const index = resourceBodyParameters.indexOf(param);
+
+        if (param['x-example']) {
```
pksunkara: Can't you make this logic a separate function since we are duplicating it?

src/parser.js
```diff
@@ -918,29 +965,15 @@ export default class Parser {
   }
 
   // Generates body asset from formData parameters.
-  generateBodyFromFormParameters(transitionParameters, formParameters, request) {
+  generateBodyFromFormParameters(transitionParameters, formParameters, schema) {
```
pksunkara: This should be named `generatedSchemaFromFormParameters` right?

klokane: I did not touch name :)

src/parser.js
klokane: Agree, but there will be next changes related to `multipart/form-data`.
Let postpone refactoring of this function once done all changes.

There is too `Path` class refactoring proposal as sugest @kylef

pksunkara: Oh, I thought this PR is done. Then I can wait until later to review.

klokane: It is done, "multipart" is separate task

klokane: There is new task in trello board

pksunkara:
@klokane Can you update changelog and make a release commit too?

klokane:
Updated Version & Chnagelog

pksunkara:
You are missing the test case for this:

```
##### Scenario: Both path and operation has same query parameter

- Given Swagger with path `/`
- When the path has `query` parameter named `test`
- When the path has operation `get`
- When the operation has `query` parameter named `test`
- Then `href` of `resource` should be `/{?test}`
- Then `href` of `transition` should not exist
```

klokane:
This one?
https://github.com/apiaryio/fury-adapter-swagger/pull/98/files#diff-3e2e7d9ca8ff0363d9346425d08383c8R161

pksunkara:
No, that's about 2 path params with same name. I am talking about 2 query params with same name.

klokane:
They have same name `test`

klokane:
One on path, second on operation

pksunkara:
But they are of `type = path` params. I am talking about `type = query` params with same name. This test case is supposed to be here. https://github.com/apiaryio/fury-adapter-swagger/pull/98/files#diff-3e2e7d9ca8ff0363d9346425d08383c8R131

klokane:
OK you are right.
I'm pretty sure I did it. It seems, I accidentally remove it.
Added, will push in few moments

PR #101, opened 2017-04-11
chore: Update fury to 3.0.0-beta.0 and minim to 0.15.0

pksunkara:


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.10.1 - 2017-04-11
+
+- Update fury to 3.0.0-beta.1
```
kylef: This patch update would introduce a breaking change of dropping node 0.10/0.12 support?

pksunkara: No, that was done already in 0.10.0

kylef: Also, it would break compatibility with older minim so this would need to be 0.11

pksunkara: What do you mean by that? The output or the API is not getting changed at all. Only the internals are.

kylef: Any dependent that uses fury/fury-adapter-swagger would break unless they are pinned to a specific version.

If I have something like this as dependencies:

```json
{
  "dependencies": {
    "fury": "^2.1.0",
    "fury-adapter-swagger": "^0.10.0"
  }
}
```

Then when I install my dependencies I'd get a version of 2 fury and 0.10.1 of the adapter which is not compatible with fury2/minim. It's a breaking change. Other fury will provide older minim which would not pass elements down (ex: `resource.href.toValue()`).

pksunkara: Oh yeah, I forgot that fury is a dev dependency here and not a normal dependency.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.11.0 - 2017-04-11
+
+- Update fury to 3.0.0-beta.1
```
kylef: It's beta 0 not beta 1? https://github.com/apiaryio/fury.js/releases/tag/v3.0.0-beta.0

PR #102, opened 2017-04-25
Pin to yaml-js ^0.1.5

kylef:


pksunkara:
@kylef Can you add a release commit too? Thanks.

PR #106, opened 2017-06-14
Ensure that x-example is used when there are items

kylef:
This PR fixes one of the bugs mentioned at https://github.com/apiaryio/dredd/issues/791 where `x-example` is ignored with array parameters where there is an `items` property.

This is because when `items` exist we are overriding the existing value, this PR changes it so that we won't use items for value if there was already `x-example`. I am not sure if this is the best fix, and maybe this could break some other behaviour I am not aware of.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1194,11 +1194,13 @@ export default class Parser {
         element.attributes.set('samples', e);
       }
     } else {
-      element = new Type();
+      let example;
 
       if (parameter['x-example'] !== undefined) {
-        element.content = parameter['x-example'];
+        example = parameter['x-example'];
       }
+
+      element = new Type(example);
     }
 
     // If there is a default, it is set on the member value instead of the member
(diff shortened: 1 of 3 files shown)


src/parser.js
```diff
@@ -1194,11 +1194,13 @@ export default class Parser {
         element.attributes.set('samples', e);
       }
     } else {
-      element = new Type();
+      let example;
 
       if (parameter['x-example'] !== undefined) {
-        element.content = parameter['x-example'];
```
kylef: Note, if we set the `.content` of an element, when the new value is an array, the array has to be mimim elements. Whereas when we pass the value to the constructor of an Array element it will refract all of the given items in the array.

frockert:
When will this be release? We are hitting this problem which is blocking us to have the validations running. Can we release it ASAP please? 
Thanks

kylef:
@frockert Hey, we're working on the releasing this fix across the entire product. There are some changes we need to make on the Apiary side before we can fully get this out unfortunately, but we're working on it.

frockert:
@kylef  do you have a timeline for that?

kylef:
@frockert I apologise for not getting back to you, this has been deployed for a little while now and available in Apiary.

PR #107, opened 2017-06-14
Support csv and multi collectionFormat for query parameters

kylef:
Add supporting for `csv` and `multi` collectionFormat in query parameters. `ssv`, `tsv`, `pipes` are still not supported, that is because they do not translate directly into a URI Template so would involve additional work. I do not believe that `ssv`, `tsv` nor `pipes` are widely used and thus I do not think this is that much of a problem in reality, I do not think I've ever seen an API use such parameter styles. 

https://github.com/apiaryio/fury-adapter-swagger/commit/7930c444e320544b0d76e5d1c22bf7bd5470e4ab contains the refactoring necessarily to fix the problem in https://github.com/apiaryio/fury-adapter-swagger/commit/8b084ae5b1e04d10ad780bcac810e554e1c3c237.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Master
+
+- Adds support for `csv` and `multi` parameter collectionFormat for query
+  parameters.
+
 # 0.12.0-beta.2
 
 ## Bug Fixes
--- a/src/uri-template.js
+++ b/src/uri-template.js
@@ -1,20 +1,16 @@
 import _ from 'lodash';
 
 export default function buildUriTemplate(basePath, href, pathObjectParams = [], queryParams = []) {
-  if (queryParams.length > 0 || pathObjectParams.length > 0) {
-    // Path object parameters apply to all nested resources (operations). Only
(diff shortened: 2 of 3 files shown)


honzajavorek:
How is the example value processed? Is it left as is for Dredd to process (csv, etc.), or does the URI Template syntax ensure that?

Also, I've seen previously that when there was something unsupported, the adapter attached some sort of a warning about "data lost". I think we could do that for any format which isn't csv or multi.

kylef:
> How is the example value processed? Is it left as is for Dredd to process (csv, etc.), or does the URI Template syntax ensure that?

Yes, the explode `*` indicates how the URI Template would be expanded. See https://tools.ietf.org/html/rfc6570#section-3.2.8 for examples.

> the adapter attached some sort of a warning about "data lost". I think we could do that for any format which isn't csv or multi.

Yes that would be great, but given the architecture of the current uri template generator it doesn't have a way to provide annotations back. This change is better that the current state so I would go for the more incremental approach.

honzajavorek:
Cool.

PR #109, opened 2017-06-16
Escape generated URI Template variables

kylef:
URI Template variables may only contain `ALPHA / DIGIT / "_" / pct-encoded` as per the URI Template specification (RFC 6570 section 2.3):

> 2.3.  Variables
> 
>      variable-list =  varspec *( "," varspec )
>      varspec       =  varname [ modifier-level4 ]
>      varname       =  varchar *( ["."] varchar )
>      varchar       =  ALPHA / DIGIT / "_" / pct-encoded

As such, when we generated URI Template variable names they should be correctly escaped as pointed out in https://github.com/apiaryio/dredd/issues/799. This pull request adds this escaping.

### DIFF
--- a/src/uri-template.js
+++ b/src/uri-template.js
@@ -1,26 +1,29 @@
 import _ from 'lodash';
 
+function escapeUriTemplateVariable(variable) {
+  return encodeURIComponent(variable)
+    .replace(/[-.!~*'()]/g, c => `%${c.charCodeAt(0).toString(16)}`);
+}
+
 export default function buildUriTemplate(basePath, href, pathObjectParams = [], queryParams = []) {
   const parameterNames = _.chain(pathObjectParams)
     .concat(queryParams)
     .filter(parameter => parameter.in === 'query')
     .uniqBy(parameter => parameter.name)
     .map((parameter) => {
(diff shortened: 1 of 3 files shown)


src/uri-template.js
```diff
@@ -1,26 +1,31 @@
 import _ from 'lodash';
 
+function escapeUriTemplateVariable(variable) {
+  return encodeURIComponent(variable)
+    .replace(/[\-\.!~*'()]/g, (c) => {
```
kylef: `encodeURIComponent` will only do half of the job and leaves `alphabetic, decimal digits, - _ . ! ~ * ' ( )` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) unescaped so this code escapes the remaining characters so that we escape everything except alphanumeric, decimal digits and `_`.

pksunkara:
Looks good but tests failing.

src/uri-template.js
honzajavorek: Smart!

kylef:
@pksunkara I've fixed the code linting failure.

PR #116, opened 2017-08-01
fix: Add warning when x-example is not of correct type

pksunkara:
If you look at the validation for defaultValue, we do not care about number and boolean validations. So, I haven't done it for sample value too. I have only fixed the array validation for now because there are other validations we want to do when we want to start on it. It's better as a separate task.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1147,6 +1147,7 @@ export default class Parser {
 
     let element;
     let Type;
+    let example = parameter['x-example'];
 
     // Convert from Swagger types to Minim elements
     if (parameter.type === 'string') {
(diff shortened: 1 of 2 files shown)


src/parser.js
```diff
@@ -1157,12 +1158,19 @@ export default class Parser {
       Type = BooleanElement;
     } else if (parameter.type === 'array') {
       Type = ArrayElement;
+
+      if (example && !Array.isArray(example)) {
```
kylef: The check is only for array types, I think we could make this generic and support all types. What do you think?

Mostly (possibly working) pseudo code but you get the idea:

```js
const parameterTypes = {
  string: {
    element: StringElement,
    type: String
  },
  array: {
    element: ArrayElement,
    type: Array
  },
  // all types go here
}

const parameterType = parameterTypes[parameter.type];
if (parameterType) {
  const element = new parameterType.element();

  if (example) {
    if (example instance parameterType.type) {
      element.content = example;
      // todo set source map?
    } else {
      // create annotation of example being wrong type
    }
  }
} else {
  // create annotation of unsupported/unknown type
}
```

kylef:
@pksunkara Can you add a CHANGELOG entry please.

pksunkara:
That's what I wrote in the description. It's not just limited to sample value. We need to do the checks elsewhere too. Which is why it's better as a separate task.

pksunkara:
Re: changelog, I wanted to leave it to Thomas as part of his release task. And then introduce him about making it part of PRs.

PR #119, opened 2017-08-11
Handle parameter array types when enumeration is provided

kylef:
In Swagger a [Parameter Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#parameter-object) may contain enum which should match the same the parameter. For example, if I declare an array parameter which the items of the array would be strings. I would have an enumeration of arrays of strings which is permitted in specification. Any type MAY be inside the enum array [[spec](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.5.1)]:

> **Elements in the array MAY be of any type, including null.**

While, I don't think it makes logical sense to do so. It is permitted in specification and I am seeing people do this in Swagger. Before this PR we was setting an elements content to an array which contained an array of primitive JavaScript types. Since an elements content can contain an array of elements, but not an array of primitive types the serialiser was failing for these documents. Correctly refracting the value would solve this problem.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1183,8 +1183,7 @@ export default class Parser {
       const enumerations = new ArrayElement();
 
       _.forEach(parameter.enum, (value, index) => {
-        const e = new Type();
-        e.content = value;
+        const e = new Type(value);
 
         if (this.generateSourceMap) {
           this.createSourceMap(e, path.concat('enum', index));
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Master
+# 0.13.1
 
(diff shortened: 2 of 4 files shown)


src/parser.js
```diff
@@ -1183,8 +1183,7 @@ export default class Parser {
       const enumerations = new ArrayElement();
 
       _.forEach(parameter.enum, (value, index) => {
-        const e = new Type();
-        e.content = value;
+        const e = new Type(value);
```
kylef: This fix is pretty trivial. That's because the constructor for an ArrayElement will correctly refract the given values inside the array. Whereas the setter for `e.content` does not handle this.

PR #120, opened 2017-08-17
Prevent re-use of auth scheme elements

kylef:
Fixes https://github.com/apiaryio/dredd-transactions/issues/93

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -778,7 +778,7 @@ export default class Parser {
           // refactor the code below as this is a little weird.
           relevantResponses.null = {};
         } else {
-          this.createTransaction(transition, method, schemes);
+          this.createTransaction(transition, method, schemes.map(element => element.clone()));
         }
       }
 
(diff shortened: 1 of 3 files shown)


src/parser.js
```diff
@@ -778,15 +778,15 @@ export default class Parser {
...
+          this.createTransaction(transition, method, schemes.map(element => element.clone()));
         }
       }
 
       // Transactions are created for each response in the document
       _.forEach(relevantResponses, (responseValue, statusCode) => {
         this.handleSwaggerResponse(transition, method, methodValue,
                                    transitionParams, responseValue, statusCode,
-                                   schemes, resourceParams);
+                                   schemes.map(element => element.clone()), resourceParams);
```
kylef: Since we're passing in same schemes array for each iteration of this loop (and also above) we would pass same schemes which would be re-used.

antkazam:
+1 looking forward to getting this fixed! Thanks!

PR #124, opened 2017-08-24
Validate and discard invalid parameter values

kylef:
Parameter example values, default values and enumeration values are now validated against the parameter type and items. Any invalid parameter value will cause a warning and be discarded.

The following is now validated:

- Example Value
- Default Value
- Enumeration Values

The following information is used for validations:

- Parameter Type
- Parameter Items, when the type is an array, the contents of the array will now be validated too.

The first commit in this PR refactors converting a parameter to consistently use `this.path` like the rest of the code base instead of a passed path. This makes it easier to implement the feature in the following commit. The final commit in the PR removes a redundant piece of code.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -1188,6 +1188,13 @@ export default class Parser {
       validator.getLastError().details.forEach((detail) => {
         this.createAnnotation(annotations.VALIDATION_WARNING, this.path, detail.message);
       });
+
+      // Coerce parameter to correct type
+      if (schema.type === 'string') {
+        if (typeof value === 'number' || typeof value === 'boolean') {
+          element = new this.minim.elements.String(String(value));
+        }
+      }
     }
 
     return element;
--- a/src/parser.js
+++ b/src/parser.js
@@ -1272,10 +1272,6 @@ export default class Parser {
       if (parameter.required) {
         element.attributes.set('typeAttributes', ['required']);
       }
-
-      if (parameter.default !== undefined) {
-        element.attributes.set('default', parameter.default);
-      }
     }
 
     return element;
--- a/src/parser.js
+++ b/src/parser.js
@@ -4,6 +4,7 @@ import _ from 'lodash';
 import yaml from 'js-yaml';
 import typer from 'media-typer';
 import SwaggerParser from 'swagger-parser';
+import ZSchema from 'z-schema';
 import annotations from './annotations';
 import { bodyFromSchema, bodyFromFormParameter } from './generator';
 import uriTemplate from './uri-template';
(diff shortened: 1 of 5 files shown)


test/parameter.js
```diff
@@ -135,8 +74,12 @@ describe('Parameter to Member converter', () => {
     const member = parser.convertParameterToMember(parameter);
 
     expect(member.value).to.be.instanceof(minim.elements.Array);
-    expect(member.value.toValue()).to.deep.equal([]);
-    expect(parser.result.toValue()).to.deep.equal(['Value of example should be an array']);
+
+    expect(member.value.length).to.equal(1);
```
kylef: Because of the example value was being set (and not discarded) in the previous implementation. A further check where we determine if there is no values and set an empty value from the parameter items type wasn't previously handled:

```js
if (parameter.type === 'array' && parameter.items && !example) {
  element.content = [this.convertParameterToElement(parameter.items, true)];
}
```

src/parser.js
```diff
@@ -1138,132 +1139,164 @@ export default class Parser {
...
 
-    const sampleContent = new Type(example);
+  createAnnotationsFromSchemaValidation(error) {
+    error.details.forEach((detail) => {
+      this.createAnnotation(annotations.VALIDATION_WARNING, this.path, detail.message);
+    });
+  }
+
+  // Convert a Swagger parameter into a Refract element.
+  convertParameterToElement(parameter, setAttributes = false) {
```
kylef: This function is pretty long, it was before I started. I have made some improvements to the function layout, it is in three main sections and the isolation between them is further.

- Example
- Enum
- Default
- Array value from items if no example
- setAttribute

Each section really only knows about the parameter, validation of value and element now.

test/parameter.js
```diff
@@ -177,4 +120,55 @@ describe('Parameter to Member converter', () => {
     expect(enumerations).to.be.instanceof(minim.elements.Array);
     expect(enumerations.toValue()).to.deep.equal([['hello']]);
   });
+
+  it('creates a warning when example does not match parameter type', () => {
+    const parser = new Parser({ minim, source: '' });
+    parser.result = new minim.elements.ParseResult();
+    parser.convertParameterToMember({
+      type: 'string',
+      'x-example': 5,
```
honzajavorek: So this would result in dropping the parameter? If people were relying in numbers being coerced to strings, this could result in quite a breaking change.

honzajavorek:
I'd be cautious about dropping the parameters, but if they're invalid, then it's probably correct behavior. Sometimes I'd be maybe more benevolent - I guess when people specify string and put number as example, there's definitely mismatch, but I'd go with warning + coercing the value instead of dropping the parameter altogether. Otherwise I think this should not affect Dredd much.

test/parameter.js
kylef: Previously they were not coerced as strings. It would become an invalid element, for example an array element that had a content of string. For the following example:

```yaml
type: array
items:
  type: number
x-example: hello
```

I don't think we can do much here other than drop the `hello` it is not an array and it is certainly not a number to go inside an array. The user was given a warning and they should know how to fix the problem.

kylef: Some of these combinations were causing exceptions to raised in some code paths.

kylef:
I think what you propose can be done, but I do not think it would be simple. We wouldn't be able to coerce in a lot of combinations, the only things I think we can safely convert is number/boolean to string. I've made #126 with a start, do you have some other suggestions on types that can be safely coerced?

honzajavorek:
I can't think of any other, actually. Coercing strings which look like numbers would be like PHP, I'm not sure we want to go that way. I believe having string as type and number as an example can be a common mistake, so this would be about smoother user experience.

PR #130, opened 2017-09-06
Support data structure referencing instead of using dereferenced version

kylef:
This is an incomplete pull request to support referenced data structures.

This has various benefits:

- Adds support for circular referenced schemas
- Decent performance boost when there are lots of referenced schemas (seeing ~10 seconds shaved on a ~27 second Swagger document)
- Adds data structure section
- DRYies up the resultant parse result significantly

### Remaining Tasks

- [ ] Changelog
- [ ] https://github.com/apiaryio/fury-adapter-swagger/pull/130#discussion_r137386128
- [ ] Test ref in all cases
    - [ ] `allOf` needs additional support
    - [ ] `oneOf`

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -106,7 +106,12 @@ export default class Parser {
       },
     };
 
-    return swaggerParser.validate(loaded, swaggerOptions, (err) => {
+    // Swagger parser is mutating the given input and dereferencing.
+    // Let's give it no changes to screw the original and give it a deep copy
+    const swaggerParserInput = JSON.parse(JSON.stringify(loaded));
+    this.loaded = loaded;
+
+    return swaggerParser.validate(swaggerParserInput, swaggerOptions, (err) => {
       const swagger = swaggerParser.api;
       this.swagger = swaggerParser.api;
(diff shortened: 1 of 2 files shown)


src/parser.js
```diff
@@ -1392,9 +1430,11 @@ export default class Parser {
     }
 
     if (handledSchema) {
+      // can we find the schema?
```
kylef: Needs to be fixed, in the case that the response/request was referenced `this.path` will be incorrect from dereferenced document. I think it will be a hassle to solve this, we need to deref document except the schemas. Workaround could be attempt to fix it and fallback to derefed schema.

src/schema.js
```diff
@@ -182,7 +183,10 @@ export default class DataStructureGenerator {
 
     let element;
 
-    if (schema.enum) {
+    if (schema['$ref']) {
+      const id = schema['$ref'].replace('#/definitions/', '');
```
kylef: 😬

pksunkara:
Looks good as a POC

PR #133, opened 2017-10-03
Added multiple produces and consumes support

pksunkara:
#### Dependencies

- [ ] Merge and Release apiaryio/swagger-zoo#45
- [ ] Rebuild CI

PR #134, opened 2017-10-03
Clean up the request/response pair generation

### LINKED ISSUE #96: Having multipart/form-data is generating urlencoded data
Given the following swagger document, even though the consumes mentions `multipart/form-data`, the generated request asset has content type `application/x-www-form-urlencoded` and url encoded data which is not what the user is requesting.

```yaml
swagger: '2.0'
info:
  title: Form Data
  version: '1.0'
consumes:
  - multipart/form-data
paths:
  '/test':
    post:
      parameters:
        - name: name
          in: formData
          required: true
          type: string
      responses:
        200:
          description: 'My Response'
```

kylef:
This pull request is similar to #133 however, however it ensures that unknown and other produces/consume content-types such as `application/x-www-form` etc are created correctly. The loops inside #133 only include JSON types which drops non-JSON content types. These changes also fix #96.

The following problems are resolved:

- `Accept` headers are present for responses which have examples, previously accept was missing here.
- Multiple JSON consumes/produces are supported. Previously only the first content type was used and the rest was discarded.
- `multipart/form-data` consumes content type was discarded and replaced by`application/x-www-form` (#96)
- Other content types are not discarded.

A full breakdown of the changes can be found in https://github.com/apiaryio/swagger-zoo/pull/46

### Dependencies

- [ ] https://github.com/apiaryio/swagger-zoo/pull/46 - This PR is required to be merged deployed to make the tests pass.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -30,6 +30,15 @@ function isExtension(value, key) {
   return _.startsWith(key, 'x-');
 }
 
+function isValidContentType(contentType) {
+  try {
+    typer.parse(contentType);
+  } catch (e) {
+    return false;
+  }
+  return true;
+}
+
 function isJsonContentType(contentType) {
   try {
     const type = typer.parse(contentType);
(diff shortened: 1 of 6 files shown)


src/parser.js
```diff
@@ -979,17 +1023,16 @@ export default class Parser {
       }
 
       if (contentType) {
-        this.withPath('examples', contentType, () => {
-          pushHeader('Content-Type', contentType, response, this);
-        });
+        if (responseValue.examples && responseValue.examples[contentType]) {
```
kylef: This should just check if `responseBody` is undefined.

PR #136, opened 2017-10-09
Allow generating example request bodies for multipart/form-data

kylef:
Request example bodies are now generated for formData parameters while using the `multipart/form-data` consumes content type.

#### Dependencies

- [x] https://github.com/apiaryio/swagger-zoo/pull/47

### DIFF
--- a/src/generator.js
+++ b/src/generator.js
@@ -3,6 +3,7 @@ import querystring from 'querystring';
 import faker from 'json-schema-faker';
 import annotations from './annotations';
 import { inferred } from './link';
+import { isFormURLEncoded, isMultiPartFormData, parseBoundary } from './media-type';
 
 faker.option({
   useDefaultValue: true,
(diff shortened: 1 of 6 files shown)


src/parser.js
```diff
@@ -53,6 +48,33 @@ function isJsonContentType(contentType) {
   }
 }
 
+function isMultiPartFormData(contentType) {
+  try {
+    const type = typer.parse(contentType);
+    return type.type === 'multipart' || type.subtype === 'form-data';
```
pksunkara: Shouldn't this be `&&`? Similarly below.

src/parser.js
```diff
@@ -882,12 +904,21 @@ export default class Parser {
...
     this.withPath(() => {
       const consumeIsJson = contentType && isJsonContentType(contentType);
+      const consumeIsMultipartFormData = contentType && isMultiPartFormData(contentType);
+
+      if (consumeIsMultipartFormData && !hasBoundary(contentType)) {
+        // When multipart/form-data conntent type doesn't have a boundary
+        // add a default one `md5('boundary')` which hopefully isn't found
+        // in an example content
+        contentType += '; boundary=e4ef8847482d240d0a3ac612876ba8d1';
+      }
```
pksunkara: I think this boundary generation should be done in generator.js

src/parser.js
kylef: Yes, I've fixed this. The problem exists in another method too which I fixed, I copied the similar code and adapted for multipart.

src/generator.js
```diff
@@ -22,9 +32,17 @@ export function bodyFromSchema(schema, payload, parser, contentType = 'applicati
         // Form data
         // TODO: check for arrays etc.
         body = querystring.stringify(body);
-      } else if (contentType === 'multipart/form-data') {
-        // TODO: Unimplemented
-        return null;
+      } else if (type && type.type === 'multipart' && type.subtype == 'form-data') {
```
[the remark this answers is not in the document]
kylef: It was in separate file, so I've moved the logic into `media-type.js` which both files import and use now.

kylef:
I've updated this to do the following:

- Changed default boundary from `md5(boundary)` to just `BOUNDARY` as I think this behaviour is clearer than an obscure hash, however it does increase the risk that the boundary clashes with some example values or keys.
- Moved all media-type specific logic to separate file `media-type.js`

pksunkara:
Looks good. Didn't check why the tests are failing though.

PR #143, opened 2017-11-05
Don't build request/response example pairs for all produces types

kylef:
This changes the behaviour of the request/response example pair generation for Swagger 2.0 so that it will not always use all of the produces content types while generating response examples. The new behaviour is as follows:

- Use ALL examples content types for response generation
- If no examples, use ONLY first JSON content type
- Fallback to empty response

This aims at solving various problems where response pair generation for certain content types is not intended. Here's an example:

```yaml
produces:
- application/json
- application/problem+json
paths:
  /:
    get:
      responses:
        200:
          schema:
            type: object
        400:
          examples:
            application/problem+json
```

In this case, I will have two request/response pairs. The first response will be of 200 status code using the first JSON content type which is `application/json`. Since the second response (400 status code) has an explicit examples section, no produces will be used during response generation and only the explicit content type `application/problem+json` will be used.

The downside to this is that if I have multiple content types in produces for example `application/json` and `text/json` and I'd like to have response pairs for each types this would no longer be generated unless I explicitly added example response bodies for both types.

#### Dependencies

- [ ] https://github.com/apiaryio/swagger-zoo/pull/51

I'd appreciate if both @honzajavorek could take a look at this PR in consideration of Dredd and the mock server, and @zdne to take a look to see if this works for all his cases.

### DIFF
--- a/test/parser.js
+++ b/test/parser.js
@@ -1,0 +1,77 @@
+import { expect } from 'chai';
+import fury from 'fury';
+import Parser from '../src/parser';
+
+describe('Parser', () => {
+  let parser;
+
+  before(() => {
+    parser = new Parser({ minim: fury.minim });
+    parser.swagger = {
+      consumes: [],
+      produces: [],
+    };
+  });
+
+  context('content types', () => {
+    it('gathers null response type when no examples or produces', () => {
+      const methodValue = {};
+      const examples = {};
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal([null]);
+    });
+
+    it('gathers all example response content types', () => {
+      const methodValue = {};
+      const examples = {
+        'application/json': '',
+        'application/hal+json': '',
+        'application/xml': '',
+      };
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal([
+        'application/json',
+        'application/hal+json',
+        'application/xml',
+      ]);
+    });
+
+    it('gathers first JSON produces without examples', () => {
+      const methodValue = {
+        produces: [
+          'application/json',
+          'application/hal+json',
+        ],
+      };
+      const examples = {};
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal(['application/json']);
+    });
+
+    it('gathers only example response content types with produces', () => {
+      const methodValue = {
+        produces: [
+          'application/json',
+          'application/problem+json',
+        ],
+      };
+      const examples = {
+        'application/vnd.error+json': '',
+      };
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal(['application/vnd.error+json']);
+    });
+
+    it('rejects invalid content types when gathering response content types', () => {
+      const methodValue = {};
+      const examples = { '!!!': '' };
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal([null]);
+    });
+  });
+});
--- a/test/parser.js
+++ b/test/parser.js
@@ -41,6 +41,7 @@ describe('Parser', () => {
     it('gathers first JSON produces without examples', () => {
       const methodValue = {
         produces: [
+          'text/plain',
           'application/json',
           'application/hal+json',
         ],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Master
+# 0.15.0
 
 ## Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Enhancements
 
 - Sample values are now generated for data structures of object and array types.
+- Request/Response pairs are now generated from explicit examples, or the first
+  JSON produces content-type.
 
 # 0.14.3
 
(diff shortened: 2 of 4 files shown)


honzajavorek:
So the example Swagger document behaves completely the same in following two cases, am I correct?

```yaml
produces:
- application/json
- application/problem+json
```

```yaml
produces:
- application/json
```

Because any other then the first `produces` gets ignored unless explicitly specified with the response example.

kylef:
@honzajavorek Correct

honzajavorek:
You mention _first JSON content type_ everywhere. Does that mean if the first `produces` is `text/plain`, it is skipped until explicitly specified with the response example? How much are non-JSON responses supported now?

kylef:
@honzajavorek That is correct. There is no behaviour for example generation for non JSON content types in produces. Produces without examples of non-JSON types are ignored. This could be changed in future though, generation of `application/x-www-form..` etc for produces would be relatively simple change if desired. We do similar generations for `consumes` request bodies.

test/parser.js
```diff
@@ -0,0 +1,77 @@
...
+    };
+  });
+
+  context('content types', () => {
+    it('gathers null response type when no examples or produces', () => {
+      const methodValue = {};
+      const examples = {};
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal([null]);
```
honzajavorek: I haven't inspected the subsequent code, but an interface using an array containing a `null` to signalize there are no content types looks weird. Wouldn't an empty array be better?

honzajavorek:
@kylef But this is only about `produces/consumes`, right? If I put non-JSON type to `examples`, it will work, no? With JSON recognized, forms recognized, and other things left as provided (text, html?).

test/parser.js
```diff
@@ -0,0 +1,77 @@
...
+          'application/hal+json',
+        ],
+      };
+      const examples = {};
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal(['application/json']);
+    });
+
+    it('gathers only example response content types with produces', () => {
```
honzajavorek: This test sounds like if the inputs were

```js
const methodValue = {
  produces: [
    'application/json',
    'application/problem+json',
  ],
};
const examples = {
  'application/hal+json': '',
  'application/vnd.error+json': '',
};
```

then following would be expected:

```
['application/vnd.error+json']
```

I'm inferring that from the word "only". If that's correct, I'd add one more content type to the examples for clarity.

test/parser.js
```diff
@@ -0,0 +1,77 @@
...
+      };
+      const examples = {
+        'application/vnd.error+json': '',
+      };
+      const contentTypes = parser.gatherResponseContentTypes(methodValue, examples);
+
+      expect(contentTypes).to.deep.equal(['application/vnd.error+json']);
+    });
+
+    it('rejects invalid content types when gathering response content types', () => {
```
honzajavorek: Should this add a warning about invalid content type in the API description? 🤔

kylef:
@honzajavorek that is also correct. If there are explicit examples they will always be used, regardless of the content-type.

test/parser.js
```diff
@@ -0,0 +1,77 @@
...
+        'application/json',
+        'application/hal+json',
+        'application/xml',
+      ]);
+    });
+
+    it('gathers first JSON produces without examples', () => {
+      const methodValue = {
+        produces: [
+          'application/json',
```
honzajavorek: I'd prepend the array with one non-JSON content type so it's clear those would be skipped.

test/parser.js
honzajavorek: Also I'm not sure if I'd like such behavior as it seems a bit counter-intuitive to me. Generally it seems the `examples` have higher priority as a source of truth about content types and this would mean they're sort of influenced by `produces`, not overriding it.

honzajavorek:
(Also, tests failed but I guess that's because of zoo...)

test/parser.js
kylef: That's already handled elsewhere. It is not handled in this specific unit.

- https://github.com/apiaryio/fury-adapter-swagger/blob/master/src/parser.js#L178
- https://github.com/apiaryio/fury-adapter-swagger/blob/master/src/parser.js#L712

test/parser.js
kylef: Added in c86d5bc

test/parser.js
kylef: These content types are mapped directly into request/response. `null` represents an empty content type and thus simplifies code elsewhere. There is code similar to:

```javascript
responseContentTypes.forEach((responseContentType) => {
  requestContentTypes.forEach((requestContentType) => {

  })
})
```

Where if either content-type is null it just represents a request/response without a content type. Perhaps the `ContentTypes` component of the method name could be clearer. These methods are generating all of the possible request/response pairs and their respective content type.

test/parser.js
kylef: I intended the test title (`gathers only example response content types with produces`) as the following:

Gather only example response content types (with produces), i.e, there is some produces too. I can see how this can be unclear, but not sure how to better explain it in such a short amount of test. Do you have some suggestion?

test/parser.js
honzajavorek: I see. Makes sense!

test/parser.js
honzajavorek: Wait, my first reading was wrong. My head mixed `application/problem+json` and `application/vnd.error+json`, thinking it's the same thing. Please let me re-review this part in the context of such a surprising finding 😄

honzajavorek: @kylef I think when I corrected the neuron connections in my brain, the description feels much more fine then previously. I think it's good to go.

PR #148, opened 2017-11-14
Support type: file parameters

kylef:
#### Dependencies

- [ ] https://github.com/apiaryio/swagger-zoo/pull/53


pksunkara:
Can you explain a little bit about why removing this line works?

PR #149, opened 2017-11-24
Samples & Default of Enum

pksunkara:


test/schema.js
```diff
@@ -713,6 +713,59 @@ describe('JSON Schema to Data Structure', () => {
...
+    expect(enumerations.get(1)).to.be.instanceof(StringElement);
+    expect(enumerations.getValue(1)).to.equal('two');
+
+    const samples = dataStructure.content.attributes.get('samples');
+
+    expect(samples.length).to.equal(2);
+    expect(samples.get(0).element).to.be.equal('enum');
+    expect(samples.get(0).toValue()).to.be.deep.equal('one');
+    expect(samples.get(1).element).to.be.equal('enum');
+    expect(samples.get(1).toValue()).to.be.deep.equal('two');
```
kylef: I think it would be worth adding an additional check that `samples.get(1)`'s content is actually a string element vs a string directly. Since `toValue()` will return the string value regardless.

test/schema.js
```diff
@@ -713,6 +713,59 @@ describe('JSON Schema to Data Structure', () => {
...
+    expect(enumerations.length).to.equal(2);
+    expect(enumerations.get(0)).to.be.instanceof(StringElement);
+    expect(enumerations.getValue(0)).to.equal('one');
+    expect(enumerations.get(1)).to.be.instanceof(StringElement);
+    expect(enumerations.getValue(1)).to.equal('two');
+
+    const defaultElement = dataStructure.content.attributes.get('default');
+
+    expect(defaultElement.element).to.be.equal('enum');
+    expect(defaultElement.toValue()).to.be.deep.equal('one');
```
kylef: Ditto https://github.com/apiaryio/fury-adapter-swagger/pull/149/files#r153001597

src/schema.js
```diff
@@ -202,16 +214,31 @@ export default class DataStructureGenerator {
         element.description = new StringElement(schema.description);
       }
 
-      if (schema.default !== undefined && !_.isArray(schema.default) &&
-          !_.isObject(schema.default)) {
+      let def = schema.default;
+
+      if (def !== undefined && !_.isArray(def) && !_.isObject(def)) {
         // TODO Support defaults for arrays and objects
```
kylef: I think we should be able to remove this TODO and `isArray`/`isObject` check now, but we can do this in separate PR.

src/schema.js
```diff
@@ -202,16 +214,31 @@ export default class DataStructureGenerator {
         element.description = new StringElement(schema.description);
       }
 
-      if (schema.default !== undefined && !_.isArray(schema.default) &&
-          !_.isObject(schema.default)) {
+      let def = schema.default;
```
kylef: Can we call this `default` not `def` to remove the abbreviation? Clearer during reading. As a Pythonist `def` is shorthand for `define` too so doesn't feel right.

src/schema.js
```diff
@@ -202,16 +214,31 @@ export default class DataStructureGenerator {
...
       }
 
+      let samples = [];
+
       if (schema.examples) {
-        element.attributes.set('samples', schema.examples);
+        samples = schema.examples;
       } else if (schema.example) {
-        element.attributes.set('samples', new ArrayElement([schema.example]));
+        samples = new ArrayElement([schema.example]);
```
kylef: In other cases above the `samples` will be a JavaScript array instead of an `ArrayElement`. Can we make this consistent? The `ArrayElement` will get converted back to a normal array due to the `map` below too.

Little bit more confusing when `samples` can be either an array or `ArrayElement` and makes it easier to introduce bugs at a later date when it could be either types.

src/schema.js
pksunkara: `default` is reserved. Can't use it.

PR #155, opened 2018-01-10
Coerce x-summary to string if not string

kylef:
Coerce a resource `x-summary` value to a string if it is not already a string. When a user enters an incorrect type such as boolean, number or array. The title would become an incorrect type and can cause subsequent tooling such as Metamorphoses to fail.

### DIFF
--- a/test/fixtures/x-summary-type.yaml
+++ b/test/fixtures/x-summary-type.yaml
@@ -1,0 +1,17 @@
+swagger: '2.0'
+info:
+  title: Resource summary as incorrect type coerces to string
+  version: '1.0'
+paths:
+  '/boolean':
+    x-summary: true
+    get:
+      responses:
+        200:
+          description: 'My Response'
+  '/number':
+    x-summary: 1.0
+    get:
+      responses:
+        200:
+          description: 'My Response'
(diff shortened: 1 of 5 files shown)


test/fixtures/x-summary-type.yaml
```diff
@@ -0,0 +1,17 @@
+swagger: '2.0'
+info:
+  title: Resource summary as incorrect type coerces to string
+  version: '1.0'
+paths:
+  '/boolean':
+    x-summary: true
```
kylef: `true` is a boolean literal and previously would create a BooleanElement.

test/fixtures/x-summary-type.yaml
```diff
@@ -0,0 +1,17 @@
...
+    get:
+      responses:
+        200:
+          description: 'My Response'
+  '/number':
+    x-summary: 1.0
```
kylef: `1.0` is a number literal and previously would create a NumberElement.

kylef:
https://github.com/apiaryio/fury-adapter-swagger/pull/156 should be merged first so both make it into release.

test/fixtures/x-summary-type.yaml
klokane: Agree

test/fixtures/x-summary-type.yaml
klokane: Agree

PR #158, opened 2018-01-23
Update swagger-parser to version 4

realityking:
Resolves #157 

Unfortunately test fails due to BigstickCarpet/swagger-parser#76 but I wanted to document the progress 😆

In case swagger-parser doesn't add the option, swagger-zoo would have to be changed.  Specifically this file https://github.com/apiaryio/swagger-zoo/blob/master/fixtures/features/api-elements/path-reference-typo.json and its source map.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.18.1
+
+## Enhancements
+
+- Allow `date` and `date-time` format support.
+
 # 0.18.0
 
 ## Breaking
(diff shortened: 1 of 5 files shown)


realityking:
Now that BigstickCarpet/json-schema-ref-parser#75 is merged, this should soon be unblocked. Just have to wait for things to be released and to tickle up.

pksunkara:
The json-schema-ref-parser v5 needs to be propagated into swagger-parser and then a new version needs to be released

realityking:
It's already in master of `swagger-parser`. There's some build issue according to swagger-parser#85.

Issue #112, opened 2017-07-11
Supporting x-nullable


philsturgeon:
Swagger/OpenAPI are in a weird state between v2.0 and v3.0. There have been a few keywords added to v3.0 that probably should have been backported into a v2.1 or something. Instead the authors have quite adamantly shouted "No, this is coming in v3!" and everyone has been kinda stuck waiting, or using `x-` vendor prefixes which some things support and some don't.

You folks have done an awesome job supporting some of these `x-` vendor prefixes (like `x-example`), but there is another one: `x-nullable`.

Currently you don't support `x-nullable`, which leads to Dredd pronouncing things as invalid when they are not. A workaround has been suggested in https://github.com/apiaryio/dredd/issues/283, suggesting that `type: ['string', 'null']` will work. It doesn't. 

(Maybe their renderer or parser-of-choice is lax enough to support it, but ReDoc and any strict parser will error.)

More on this here https://github.com/swagger-api/swagger-editor/issues/1302 

Instead, OpenAPI 3.0 will add `nullable: true` on the [schema object](https://github.com/OAI/OpenAPI-Specification/blob/OpenAPI.next/versions/3.0.md#fixed-fields-20). In the mean time, renderers and parsers are supporting `x-nullable: true`. 

Example: https://github.com/Rebilly/ReDoc/pull/95

Supporting this field during this odd v2.0 / v3.0 gap would be amazing, and I can't get Dredd running at work until this is done.

philsturgeon:
Updated description because I was crossing things over.

kylef:
I think it makes sense to support this, let me clear this with our product team and see if we can get this implemented soon. Thanks for the detailed report @philsturgeon.

InfopactMLoos:
Any news on this?

Marahin:
Bump, @kylef could you update us on the state of this issue?

kylef:
I'd expect most of you are asking about this to use with Dredd. Right now Dredd is receiving the Swagger specific schema so to implement this feature we would need to craft JSON Schema's so that Dredd can use a JSON schema. Or alternatively add support to Dredd for validating Swagger schemas which could include `x-nullable`.

InfopactMLoos:
@kylef, Yes I think that is what is being described in this issue. Since your first option is probably a lot of work and because this repository is specifically focused on swagger and not JSON schema I don't think it is wise to change that. 

[...]

PR #162, opened 2018-03-12
Add support for x-nullable vendored extension

kylef:
As requested in #112, to allow using null types with Dredd.


kylef:
I've removed commit 67a9be3 which is the release commit and rebased on master. I'll make separate release PR once this is merged encapsulating #161 with changelog and appropriate breaking version.

PR #164, opened 2018-03-17
Update json-schema-faker to version 0.5.0-rc13

realityking:
This requires an update to `swagger-zoo` before it'll pass the tests. See apiaryio/swagger-zoo#55.


pksunkara:
Overridden by #165

PR #177, opened 2018-05-10
support root extensions

abacaphiliac:
fix(root-extensions): similar to #174, this will parse root extensions [per the swagger specification](https://swagger.io/docs/specification/2-0/swagger-extensions/).

unfortunately, root extensions are parsed into the same location as the info extensions.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -174,6 +174,7 @@ export default class Parser {
         this.handleSwaggerInfo();
         this.handleSwaggerHost();
         this.handleSwaggerAuth();
+        this.handleSwaggerVendorExtensions(this.api, this.swagger);
 
         this.handleExternalDocs(this.api, swagger.externalDocs);
 
--- a/test/fixtures/root-extensions.json
+++ b/test/fixtures/root-extensions.json
@@ -1,0 +1,77 @@
+{
+  "element": "parseResult",
+  "content": [
+    {
+      "element": "category",
+      "meta": {
+        "classes": {
+          "element": "array",
+          "content": [
+            {
+              "element": "string",
+              "content": "api"
+            }
+          ]
+        },
+        "title": {
+          "element": "string",
+          "content": "Root Extensions"
+        }
+      },
+      "content": [
+        {
+          "element": "extension",
+          "meta": {
+            "links": {
+              "element": "array",
+              "content": [
+                {
+                  "element": "link",
+                  "attributes": {
+                    "relation": {
+                      "element": "string",
+                      "content": "profile"
+                    },
+                    "href": {
+                      "element": "string",
+                      "content": "https://help.apiary.io/profiles/api-elements/vendor-extensions/"
+                    }
+                  }
+                }
+              ]
+            }
+          },
+          "content": {
+            "x-info-extension": "Info Extension"
+          }
+        },
+        {
+          "element": "extension",
+          "meta": {
+            "links": {
+              "element": "array",
+              "content": [
+                {
+                  "element": "link",
+                  "attributes": {
+                    "relation": {
+                      "element": "string",
+                      "content": "profile"
+                    },
+                    "href": {
+                      "element": "string",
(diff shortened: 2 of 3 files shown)


pksunkara:
Can you please make a PR on swagger-zoo fixing it?

pksunkara:
You don't need to add a test case here if you fix it there.

abacaphiliac:
@pksunkara thanks for the advice! i will do that today.

abacaphiliac:
@kylef i agree with your concern.

[vendor-extensions](https://help.apiary.io/profiles/api-elements/vendor-extensions/) links to [element-definitions](http://api-elements.readthedocs.io/en/latest/element-definitions.html#extending-api-elements), which provides the following example:
```json
{
  "element": "extension",
  "meta": {
    "links": {
      "element": "array",
      "content": [
        {
          "element": "link",
          "attributes": {
            "relation": {
              "element": "string",
              "content": "profile"
            },
            "href": {
              "element": "string",
              "content": "http://example.com/extensions/info/"
            }
          }
        }
      ]
    }
  },
  "content": {
    "element": "object",
    "content": [
      {
        "element": "member",
        "content": {
          "key": {
            "element": "string",
            "content": "version"
          },
          "value": {
            "element": "string",
            "content": "1.0"
          }
        }
      }
    ]
  }
}
```

`info`, `paths`, and `root` seem like fitting `relation` `attributes` to me.

abacaphiliac:
@kylef href fragment makes sense to me unless there is another more easily parsable location.

are there any other attributes or meta that we can use with the link element? the extension element example uses an object element as content. we could put some kind of source information there, but i think that would conflict with the existing extension element content structure.

pksunkara:
I like the href fragment idea.

pksunkara:
Code looks good. Can you do a corresponding swagger-zoo PR?

src/parser.js
```diff
@@ -174,14 +174,15 @@ export default class Parser {
...
+        this.handleSwaggerVendorExtensions(this.api, this.swagger, 'root');
 
         this.handleExternalDocs(this.api, swagger.externalDocs);
 
         this.validateProduces(this.swagger.produces);
         this.validateConsumes(this.swagger.consumes);
 
         const complete = () => {
-          this.handleSwaggerVendorExtensions(this.api, swagger.paths);
+          this.handleSwaggerVendorExtensions(this.api, swagger.paths, 'paths');
```
kylef: This would be a breaking change for tooling looking for these relations so we need to be careful about rolling out this change. @pksunkara perhaps we should serialise them differently in minim 1.0/0.6 serialisers?

We have a task in our backlog about moving all of the links (such as those in annotations) to be in apielements.org (since many of them are actually 404s). What we could do it update them all and create a profile registry over in API elements. Then in the minim 0.6 serialiser we can replace them to be the existing links.

Another thing to think about is I don't think the current profile is correct given the API Elements 1.0 serialisation rules. Since we're placing an un-refracted object directly in the content of the extension which breaks the rules. Given the following extension:

```yaml
x-extension:
  element: string
```

Tooling would think that this is a string element instead a user-defined object which is some of the problems we was trying to solve in API Elements 1.0. @pksunkara Would love your thoughts, we can probably use this time to do the profile right before wide-adoption of API Elements 1.0.

kylef: An alternative idea, we could also include both the old style and new style extension element. This allows existing tooling to continue to work (we would have to think of a deprecation plan) and new tooling to directly utilise the newer format. Personally I think my other comment would be better idea long-term but it involves more work short-term (changing minim serialiser and releasing another toolchain version).

pksunkara: I like your first comment on this.

- Move the profiles to apielements.org
- Write an RFC about the content of the profiles and implement it
- Convert it into 0.6 serialization.

abacaphiliac:
@pksunkara I'd be happy to open a corresponding swagger-zoo PR.

Would you prefer I hold off on that until the profile and serializer details settle? Is there any way I can contribute to that discussion or project(s)?

PR #184, opened 2018-05-25
fixes #182: parses `info.version` as an API element attribute

### LINKED ISSUE #182: Parse info.version
Hi Team,

I need `info.version` to be available in my parse result. It looks like only `title` and `description` are parsed. I've looked into adding it as meta, like `title`. Nearly all of the swagger-zoo fixture tests fail due to the fixtures specifying a version which is apparently supposed to be ignored.

Is `version` truly supposed to be stripped, as swagger-zoo seems to indicate? ~~Can `version` be removed from the fixtures (yaml and json) so that this project can parse it additively?~~ Should I open an issue/request to swagger-zoo to add the parsed version to the fixtures? If so, is it an attribute, meta, or content?

Thanks for the consideration : )

abacaphiliac:
fixes #182: parses `info.version` as an API element attribute. breaks many tests, as `info.version` has been ignored until now.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -341,6 +341,18 @@ export default class Parser {
           });
         }
 
+        if (this.swagger.info.version) {
+          this.withPath('version', () => {
+            this.api.attributes.set('version', this.swagger.info.version);
+
+            if (this.generateSourceMap) {
+              this.createSourceMap(this.api.attributes.get('version'), this.path);
+            }
+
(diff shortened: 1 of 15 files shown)


pksunkara:
@kylef What do you think?

kylef:
This has landed in master via 4cac3665571d7f42b6bc2c60095ea8edecda6a96 and released as 0.20.0.

PR #188, opened 2018-05-28
Release 0.19.0

pksunkara:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 0.18.5
+
+## Enhancements
+
+- Enumeration behaviour is now the same as drafter such that the valus that are
```
kylef: Typo `valus`.

I'd not mention drafter but instead say it was aligned with API Elements 1.0 because a reader of this changelog may not know what drafter is and also the current released versions of drafter do not have fixed typeAttribute.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fury-adapter-swagger",
-  "version": "0.18.4",
+  "version": "0.18.5",
```
kylef: This should be 0.19.0 because the changes update the peer dependency which is not semantically permitted.

pksunkara: Updated

package.json
```diff
@@ -27,11 +27,11 @@
     "z-schema": "^3.16.1"
   },
   "peerDependencies": {
-    "fury": "3.0.0-beta.7"
+    "fury": "^3.0.0-beta.7"
```
kylef: 3.0.0-beta.8 can include semantically breaking changes so that would break 0.19.0 of fury-adapter-swagger. This should be a hard dependency.

pksunkara: Fixed.

PR #189, opened 2018-06-01
Support allOf schema without type hint

kylef:
When an object schema uses `allOf` and doesn't provide a type hint at the schema root, the `allOf` types were not matched for object sub-schemas.

These changes allow the follow schema to produce a data structure element, before it was required to insert `type: object` at the first time next to `allOf`:

```yaml
allOf:
  - type: object
    properties:
      username:
        type: string
  - type: object
     properties:
       name:
         type: string
```

### DIFF
--- a/src/schema.js
+++ b/src/schema.js
@@ -1,4 +1,4 @@
-/* eslint-disable class-methods-use-this */
+/* eslint-disable class-methods-use-this, arrow-body-style */
 
 import _ from 'lodash';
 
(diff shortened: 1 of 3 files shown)


src/schema.js
```diff
@@ -165,6 +165,26 @@ export default class DataStructureGenerator {
...
+   *
+   * In the case where there is no provided type, the allOf types are matched.
+   */
+  validateSchemaTypes(schema, type) {
+    if (schema.type === type) {
+      return true;
+    }
+
+    if (schema.type === undefined && schema.allOf && schema.allOf.length > 0) {
+      const schemasWithoutMatchingType = schema.allOf.filter((subschema) => {
```
kylef: I can't win with this line. I wanted to use what I have, but it produces a lint error:

```
/Users/kyle/Projects/apiaryio/fury-adapter-swagger/src/schema.js
  178:77  error  Unexpected block statement surrounding arrow body  arrow-body-style
```

It wants me to make this a one-liner, but then its a ridiculously long one-liner. If changed you'd then get another lint error that the line is too long. So you can't win:

```
/Users/kyle/Projects/apiaryio/fury-adapter-swagger/src/schema.js
  178:1  error  Line 178 exceeds the maximum line length of 100  max-len
```

eslint 🤦‍♂️. Anyone have a suggestion about this dumb `arrow-body-style` rule?

kylef:
Didn't add a release commit to PR because after this is merged I want to see about adding support for allOf for other types.

src/schema.js
kylef: Made eslint disable `arrow-body-style` for the file because that rule is dumb, IMO it should be down to code-write on what makes sense for given code block.

pksunkara: Can't help it since we are using airbnb as base

src/schema.js
```diff
@@ -189,7 +209,7 @@ export default class DataStructureGenerator {
       element = this.generateEnum(schema);
     } else if (schema.type === 'array') {
```
pksunkara: Why not use that here too?

PR #195, opened 2018-07-24
Prevent producing messageBody/messageSchemaBody for binary bodies

kylef:
When a user supplies `format: binary` in a request or response body root. We shouldn't generate a JSON Schema inside a messageSchemaBody as you cannot express binary in JSON Schema and thus the schema wouldn't make sense and can cause further problems such as indicated in #193. Subsequently, as you cannot express binary body examples either, no messageBody should be present.

### Dependencies

- [x] https://github.com/apiaryio/swagger-zoo/pull/60 - CI fails until this is merged/released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Return an error in the parse result when the source API Description Document
   is not an object. Previously an error was thrown.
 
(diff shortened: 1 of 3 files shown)


pksunkara:
Published swagger-zoo. Please rebase this PR and push agian.

PR #196, opened 2018-08-16
Return error when source is not an object

kylef:
Spotted this causing errors when the parser if given a non-object YAML document. An error would arise lower down when trying to set properties on for example a string.

```
   1) Swagger 2.0 adapter can parse regression fixtures Parses string

       Cannot create property 'info' on string 'Hello World'

       at Parser.parse (src/parser.js:97:7)

       96 |     if (loaded.info === undefined) {
       97 |       loaded.info = {};
       98 |     }
```

### DIFF
--- a/test/adapter.js
+++ b/test/adapter.js
@@ -27,7 +27,9 @@ function testFixture(description, fixture, generateSourceMap = false) {
       expected = fixture.apiElements;
     }
 
-    fury.parse({ source, generateSourceMap }, (err, output) => {
+    const mediaType = 'application/swagger+yaml';
+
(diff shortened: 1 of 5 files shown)


test/adapter.js
```diff
@@ -27,7 +27,9 @@ function testFixture(description, fixture, generateSourceMap = false) {
       expected = fixture.apiElements;
     }
 
-    fury.parse({ source, generateSourceMap }, (err, output) => {
+    const mediaType = 'application/swagger+yaml';
```
kylef: If we don't pass the media type for swagger to parse, the document won't be recognised as Swagger.

PR #200, opened 2018-08-23
Place example values as the element content

kylef:
The example value of a Swagger 2 Schema Object will now become the content of the element providing the following rules are met:

- There isn't already a content, for example. An array with defined items, OR an object with defined properties.

- The example value matches the type of the element.

Counter part PR: https://github.com/apiaryio/swagger-zoo/pull/62

### DIFF
--- a/src/schema.js
+++ b/src/schema.js
@@ -251,7 +251,7 @@ export default class DataStructureGenerator {
         samples = [schema.example];
       }
 
-      if (samples.length) {
+      if (samples.length > 0) {
         if (schema.enum) {
           samples = samples.map((item) => {
             const enumeration = new EnumElement(item);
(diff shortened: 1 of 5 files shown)


src/schema.js
```diff
@@ -260,7 +260,25 @@ export default class DataStructureGenerator {
...
+          // check the type matches our element. It will also refract
+          // object/array items that are the sample value as members/elements
+          // for us so we can grab its content
+
+          const example = this.minim.toElement(samples[0]);
+
+          if (element.element === example.element) {
+            element.content = example.content;
+          } else {
+            element.attributes.set('samples', samples);
```
kylef: The test case "produces string element with samples for example type mismatch" covers this branch.

src/schema.js
```diff
@@ -260,7 +260,25 @@ export default class DataStructureGenerator {
...
+          }
+        } else {
+          element.attributes.set('samples', samples);
```
kylef: The following test cases cover thsi branch:

- "produces samples from examples when we have items"
- "produces samples from examples when we have properties"

PR #203, opened 2018-08-23
Use example values for body generation

kylef:
Given the following Schema Object:

```yaml
type: object
properties:
  id:
    type: number
    example: 1
  name:
    type: string
    example: doe
```

In the past, faker would use random values for id and name. Ignoring the user-provided example value. This PR changes that so these examples are respected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Fury Swagger Parser Changelog
 
+## Master
+
+### Bug Fixes
+
+- Example values in nested Schema Object will now be used for JSON body
+  generation.
+
 ## 0.22.2 (2018-10-25)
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -19,11 +19,11 @@
   "dependencies": {
     "babel-runtime": "^6.23.0",
     "js-yaml": "^3.4.2",
-    "json-schema-faker": "0.5.0-rc13",
+    "json-schema-faker": "git+https://github.com/json-schema-faker/json-schema-faker.git#develop",
```
kylef: I am awaiting a relase for JSF, seems one will be made today/tomorrow.

kylef:
NOTE, the tests fail on CI because of the development JSF. Need to do `npm run build` in JS so some internal files are updated to support useExampleValue.

PR #204, opened 2018-08-29
Pass real JSON schema into faker

kylef:
This extracts two commits from #203 as I need them in another branch.

The changes in the commits are refactoring:

- 22df73c - Non functional refactoring, basically moves the body asset generation into one place. Where the generated JSON Schema from Schema Object is in scope for body generation which will be used in later commit.
- 7e334e0 - We now pass in the JSON Schema to Faker instead of Schema Object. Since Faker isn't designed to work with Schema Object but only JSON Schema. NOTE the JSON Schema is needed so that `examples` would be present for it in next commit.

### DIFF
--- a/src/generator.js
+++ b/src/generator.js
@@ -17,7 +17,13 @@ export function bodyFromSchema(schema, payload, parser, contentType = 'applicati
   let asset = null;
 
   try {
-    let body = schema.example || faker(schema);
+    let body;
+
+    if (schema.examples && schema.examples[0]) {
+      body = schema.examples[0];
+    } else {
+      body = faker(schema);
+    }
 
     if (typeof body !== 'string') {
       if (isFormURLEncoded(contentType)) {
--- a/src/parser.js
+++ b/src/parser.js
@@ -1487,7 +1487,7 @@ export default class Parser {
     }
 
     if (pushBody) {
-      bodyFromSchema(schema, payload, this, contentType);
+      bodyFromSchema(jsonSchema, payload, this, contentType);
     }
 
     this.pushSchemaAsset(schema, jsonSchema, payload, this.path);
--- a/src/parser.js
+++ b/src/parser.js
@@ -959,11 +959,7 @@ export default class Parser {
                 }
 
                 this.withPath('schema', () => {
-                  if (consumeIsJson) {
-                    bodyFromSchema(param.schema, request, this, contentType);
-                  }
-
-                  this.pushSchemaAsset(param.schema, request, this.path);
+                  this.pushBodyAssets(param.schema, request, contentType, consumeIsJson);
                 });
               }));
               break;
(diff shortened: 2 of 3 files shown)


src/generator.js
```diff
@@ -17,7 +17,13 @@ export function bodyFromSchema(schema, payload, parser, contentType = 'applicati
   let asset = null;
 
   try {
-    let body = schema.example || faker(schema);
+    let body;
+
+    if (schema.examples && schema.examples[0]) {
+      body = schema.examples[0];
+    } else {
+      body = faker(schema);
```
pksunkara: faker respects `schema.example`?

src/parser.js
```diff
@@ -959,11 +959,7 @@ export default class Parser {
                 }
 
                 this.withPath('schema', () => {
-                  if (consumeIsJson) {
-                    bodyFromSchema(param.schema, request, this, contentType);
-                  }
-
-                  this.pushSchemaAsset(param.schema, request, this.path);
+                  this.pushBodyAssets(param.schema, request, contentType, consumeIsJson);
```
pksunkara: Maybe rename it `pushAssets`

src/parser.js
```diff
@@ -1480,43 +1473,53 @@ export default class Parser {
...
+    }
+
+    this.pushSchemaAsset(schema, jsonSchema, payload, this.path);
+    this.pushDataStructureAsset(schema, payload);
+  }
+
+  // Create a Refract asset element containing JSON Schema and push into payload
+  pushSchemaAsset(schema, jsonSchema, payload, path) {
+    const Asset = this.minim.getElementClass('asset');
+    const schemaAsset = new Asset(JSON.stringify(jsonSchema));
```
pksunkara: Isn't this the line which throws the circular reference error? It's not getting caught.

src/parser.js
```diff
@@ -1480,43 +1473,53 @@ export default class Parser {
...
-      schemaAsset.contentType = 'application/schema+json';
-
-      if (this.generateSourceMap) {
-        this.createSourceMap(schemaAsset, path);
-      }
-
-      this.handleExternalDocs(schemaAsset, schema.externalDocs);
-
-      payload.content.push(schemaAsset);
-      handledSchema = true;
```
pksunkara: What happened to the logic this variable is being used for?

src/generator.js
kylef: No, `example` will never get here because that isn't valid in JSON Schema which is being passed in. This line will actually go away in https://github.com/apiaryio/fury-adapter-swagger/pull/203/commits/abed7a4a44d396a8ce6e6790138b2094c262f6ea once JSON Schema Faker is released.

src/parser.js
kylef: The circular reference problem / warning will be emitted before we get this far. `convertSchema` will fail and that is caught further up and warning is made and then returned so body nor schema/datastructure asset changes.

src/parser.js
kylef: There is a return on line 1486.

src/parser.js
kylef: Changed in 628e88b

src/parser.js
kylef: There ares tests for this (cirlcular-example fixture) to confirm no functionality is broken.

PR #205, opened 2018-08-30
Support circular referencing in JSON / JSON Schema

### LINKED ISSUE #45: Add recursive JSON Schema support

kylef:
This PR adds support for circular referencing. I think it is a great way to break down the work I was doing on changing how we support referencing as it is a way to introduce the reference handling into our code-base bit-by-bit. We can get this into production sooner and see how it behaves.

The JSON Schema generator fully supports referencing now, and faker already supports referencing (even circular). Data Structure will silently error out so it isn't present in parse results for circular. I will work on that next but it is a little bit more complicated and it will come into a separate PR.

### Dependencies

- [x] https://github.com/apiaryio/swagger-zoo/pull/63

Closes #45


src/json-schema.js
```diff
@@ -26,58 +31,101 @@ function convertSubSchema(schema) {
...
   }
 
   return actualSchema;
 }
 
+function lookupReference(reference, root) {
+  const parts = reference.split('/');
+
+  if (parts[0] !== '#') {
+    throw new Error('Schema reference must start with document root (#)');
```
kylef: I've tested all these error paths at the unit-level. However they shouldn't be possible in real-world because the dereferencer will catch this since only valid circular referenced will make it to lookupReference at the moment.

test/fixtures/circular-example.json
```diff
@@ -122,6 +162,66 @@
...
...                                           }\n                                                                }\n                                                              }\n                                                            }\n                                                          }\n                                                        }\n                                                      }\n                                                    }\n                                                  }\n                                                }\n                                              }\n                                            }\n                                          }\n                                        }\n                                      }\n                                    }\n                                  }\n                                }\n                              }\n                            }\n                          }\n                        }\n                      }\n                    }\n                  }\n                }\n              }\n            }\n          }\n        }\n      }\n    }\n  }\n}"
```
kylef: Just a note, faker is rendering circular JSON to a certain depth so these will always be consistently the same.

src/parser.js
```diff
@@ -1477,12 +1470,9 @@ export default class Parser {
...
     try {
-      jsonSchema = convertSchema(schema);
-    } catch (exception) {
-      this.createAnnotation(
-        annotations.DATA_LOST, this.path,
-        'Circular references in schema are not yet supported',
-      );
+      jsonSchema = convertSchema(schema, this.swagger);
+    } catch (error) {
+      this.createAnnotation(annotations.VALIDATION_ERROR, this.path, error.message);
```
kylef: The old code-path here was actually catching other errors that are not for circular referencing. For example before if there was external reference it would show the wrong error.

test/fixtures/circular-example.yaml
```diff
@@ -3,22 +3,23 @@ info:
...
 paths:
   /:
     get:
       responses:
         200:
           description: Get company info
-          examples:
-            application/json:
-              example:
-                $ref: '#/definitions/Company'
```
kylef: This wasn't valid example, looking at Swagger spec "Reference Object" is not listed inside "Example Object".

https://swagger.io/specification/v2/#exampleObject

src/json-schema.js
```diff
@@ -26,58 +31,101 @@ function convertSubSchema(schema) {
...
+  }
+
+  if (parts[1] !== 'definitions' || parts.length !== 3) {
```
pksunkara: Does it need to be definitions? This can be anything, right?

pksunkara:
My main issue is the comment I made in swagger-zoo.

src/json-schema.js
kylef: In theory, but for the Swagger document to pass the swagger schema you won't be able to place a schema in many other places.

It gets a lot more complex to support "anything" because we have to copy that reference into our schema. "definitions" is the place in Swagger 2 and also JSON Schema specification that is designed for referencing. The "definitions" keyword does not directly affect the validation result. For example, if I we're to reference some other part that could break the schema:

```
{
  "type": "array",
  "items": {
    "$ref": "#/maxItems"
  }
  "maxItems": { // placing maxItems here is invalid as maxItems has semantics in JSON Schema and shouldn't be an object.
    "type": "object"
  }
}
```

From JSON Schema spec:

> 9. Schema Re-Use With "definitions"
> 
> The "definitions" keywords provides a standardized location for schema authors to inline re-usable JSON Schemas into a more general schema. The keyword does not directly affect the validation result.
> 
> This keyword's value MUST be an object. Each member value of this object MUST be a valid JSON Schema. 

I would just support this case unless we see a others placing it elsewhere but I don't think anyone is doing that.

PR #211, opened 2018-09-05
Expose line and column in annotation source maps

kylef:
As per https://api-elements.readthedocs.io/en/latest/element-definitions.html#source-map, source maps can optionally contain a line and column number to help consumers of the parse result as these can be expensive to compute.

These are only exposed for annotation elements, we are not exposing line numbers for other elements when user requests source maps so we don't greatly enlarge the parse result size.

### Dependencies

- [x] https://github.com/apiaryio/swagger-zoo/pull/67

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -36,8 +36,8 @@ export default class Ast {
 
             if (!pieces.length) {
               // This is the last item!
-              start = subNode[0].start_mark.pointer;
-              end = subNode[1].end_mark.pointer;
+              start = subNode[0].start_mark;
+              end = subNode[1].end_mark;
             }
             break;
           }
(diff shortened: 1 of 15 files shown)


src/ast.js
```diff
@@ -52,11 +52,11 @@ export default class Ast {
           if (!newNode && piece > 0 && node.value[piece - 1]) {
             // Element in sequence does not exist. It could have been empty
             // Let's provide the end of previous element
-            start = node.value[piece - 1].end_mark.pointer;
-            end = start + 1;
```
kylef: Before we were simply guessing the end because we don't know much about the source map. This can be problematic becuase we may already be at end of file so adding one to position is incorrect, it can also be much harder for us to compute the next line/column because we'd have to traverse the source document to figure out if there is one.

I think it is simpler, safer and better to just provide the start index in this case.

PR #212, opened 2018-09-10
Don't expand optional values in circular structures

kylef:
Using `alwaysFakeOptionals` with faker and JSON Schemas with circular references is a recipe for disaster. It can lead to JSON Schema Faker getting into an infinite loop when it requires an optional value to always be faked inside a circular referenced path.

Unfortunately, removing `alwaysFakeOptionals` means that the produced example JSON bodies with circular references are less deterministic because they can now randomly fake optional or not. Once we update to json-schema-faker 0.5.0-rc15 we can set `optionalsProbability` to zero so they will never be expanded which will be a more deterministic behaviour allowing us to restore the disable test in this PR.


test/fixtures/circular-example.yaml
```diff
@@ -25,8 +26,11 @@ definitions:
         properties:
           data:
             $ref: '#/definitions/User'
+    default:
+      id: ORCL
```
kylef: We have to supply a defualt excluding the circular reference so that faker won't randomly expand optional user making the test flaky.

test/fixtures/circular-example.yaml
```diff
@@ -16,6 +16,7 @@ paths:
 
 definitions:
   Company:
+    type: object
```
kylef: Noticed the type was missing on both definitions.

PR #213, opened 2018-09-27
Add support for primitive request body types

kylef:
Adds support for primitive request body types as per https://swagger.io/docs/specification/2-0/describing-request-body/. The 3 primitive types, string number and boolean are supported.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -13,7 +13,7 @@ import { pushHeader, pushHeaderObject } from './headers';
 import Ast from './ast';
 import { DataStructureGenerator, idForDataStructure } from './schema';
 import { convertSchema, convertSchemaDefinitions } from './json-schema';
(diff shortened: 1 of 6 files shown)


src/parser.js
```diff
@@ -967,12 +969,17 @@ export default class Parser {
...
                 }
 
-                if (param.schema && param.schema.format === 'binary') {
-                  return;
+                if (param.schema) {
+                  if (param.schema.format === 'binary') {
+                    return;
+                  }
+
+                  bodyIsPrimitive = (param.schema.type && ['string', 'boolean', 'number'].includes(param.schema.type));
```
pksunkara: This should also check for `text/plain` in consumes.

kylef: Does it have to be `text/plain`? What about `text/html` or `application/my-custom-type`.

kylef:
I have to update and release swagger-zoo too so the CI passes. Let's settle review here first and then I'll go make the PR for it.

src/parser.js
pksunkara: Yeah. But, let's use the `text/*` check

PR #214, opened 2018-10-03
Generate dataStructure sections

kylef:
Adds a dataStructures section to the parse result containing every data structure found within the definitions section of a Swagger document. We now use referencing between data structures found within a request or response data structure which generally makes parse results smaller instead of duplicating the data structure contents.

Even though we are still using the referenced Swagger document for parsing, where possible I am retrieivng the schema found in the original Swagger document which hasn't been dereferenced, this won't work for every case, however it will work for most until the entire parser can handle the original document containing references.

#### Dependencies

- [ ] https://github.com/apiaryio/swagger-zoo/pull/70

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -121,7 +121,8 @@ export default class Parser {
 
     // Swagger parser is mutating the given input and dereferencing.
     // Let's give it no changes to screw the original and give it a deep copy
-    const dereferencedSwagger = JSON.parse(JSON.stringify(loaded));
+    const referencedSwagger = JSON.parse(JSON.stringify(loaded));
+    this.referencedSwagger = referencedSwagger;
 
     return swaggerParser.validate(loaded, swaggerOptions, (err) => {
       const swagger = swaggerParser.api;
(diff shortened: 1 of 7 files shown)


src/schema.js
```diff
@@ -1,6 +1,11 @@
 /* eslint-disable class-methods-use-this, arrow-body-style */
 
 import _ from 'lodash';
+import { parseReference } from './json-schema';
+
+export function idForDataStructure(reference) {
+  return `definitions/${parseReference(reference)}`;
```
kylef: One note, I am using the ID "definitions/{key}" because we are already populating other element keys with unique IDs which could clash. If a user had a swagger document has operationId it could clash with the definition key thus causing a collision. Since Swagger doesn't require that the names are unique across sections.

src/schema.js
```diff
@@ -205,7 +211,12 @@ export default class DataStructureGenerator {
 
     let element;
 
-    if (schema.enum) {
+    if (schema.$ref) {
+      // element = new RefElement(idForDataStructure(schema.$ref));
```
kylef: IMO this should be a reference element. However doing so doesn't work with attributes kit 😞.

src/parser.js
```diff
@@ -638,6 +650,38 @@ export default class Parser {
...
+            dataStructure.content.id = idForDataStructure(`#/definitions/${key}`);
+
+            if (this.generateSourceMap) {
+              this.createSourceMap(dataStructure, this.path);
+            }
+
+            dataStructures.push(dataStructure);
+          }
+        } catch (error) {
+          // TODO: Expose errors once feature is more-complete
```
kylef: This is the same TODO in other data structure generation. I'd like to remove it and produce a warning that it couldn't be generated. I will do that in separate PR for both cases.

PR #215, opened 2018-10-08
Handling Schema Referencing

kylef:
This pull request achieves a couple of things:

1) All references in schemas of source Swagger documents will be present in the JSON Schema found inside a parse result. This can compact the size of large schemas which have duplicate references as they will no longer be flattened.
2) We now work on the non-dereferenced tree if possible for converting the schema to JSON Schema which means we can prevent redundant conversions by converting all of the definitions upfront and caching the result for future use.

With one of my large test documents, I am seeing significant parse time improvements with this changeset, in particular the second commit which adds caching of the converted definitions. Previously for each time the same referenced definition was used, we would be converting it to JSON Schema. This can add up and cause a lot of redundant work to be done.

I parsed the same large document 4 times over both this branch and master and here are the timing results:

- this branch: 4s 951ms, 5s 33ms, 4s 862ms, 5s 201ms
- master: 20s 970ms, 22s 567ms, DNF (JavaScript heap out of memory), 21s 141ms

The DNF was likely caused by JSON Schema Faker producing a a collection of extremely large JSON bodies hitting the memory limits.

Counterpart PR: https://github.com/apiaryio/swagger-zoo/pull/71

### DIFF
--- a/src/json-schema.js
+++ b/src/json-schema.js
@@ -134,13 +134,74 @@ function checkSchemaHasReferences(schema) {
   });
 }
 
+/** Traverses the entire schema to find all of the references
+ * @returns array of each reference that is found in the schema
+ */
+function findReferences(schema) {
+  if (schema.$ref) {
+    return [schema.$ref];
+  }
+
+  let references = [];
+
+  if (schema.allOf) {
+    references = references.concat(...schema.allOf.map(findReferences));
+  }
+
+  if (schema.anyOf) {
+    references = references.concat(...schema.anyOf.map(findReferences));
+  }
+
+  if (schema.oneOf) {
+    references = references.concat(...schema.oneOf.map(findReferences));
+  }
+
+  if (schema.not) {
+    references = references.concat(...findReferences(schema.not));
+  }
+
+  // Array
+
+  if (schema.items) {
+    if (Array.isArray(schema.items)) {
+      references = references.concat(...schema.items.map(findReferences));
+    } else {
+      references = references.concat(findReferences(schema.items));
+    }
+  }
+
+  if (schema.additionalItems && typeof schema.additionalItems === 'object') {
+    references = references.concat(findReferences(schema.additionalItems));
+  }
+
+  // Object
+
+  if (schema.properties) {
+    Object.keys(schema.properties).forEach((key) => {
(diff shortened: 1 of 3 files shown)


src/json-schema.js
```diff
@@ -134,13 +134,74 @@ function checkSchemaHasReferences(schema) {
   });
 }
 
+/** Traverses the entire schema to find all of the references
+ * @returns array of each reference that is found in the schema
+ */
+function findReferences(schema) {
```
kylef: This method is pretty identical to the above `convertSubSchema` function, instead of converting. It simply returns all of the references found in the schema.

src/json-schema.js
```diff
@@ -134,13 +134,74 @@ function checkSchemaHasReferences(schema) {
...
+  }
+
+  return references;
+}
+
 /** Convert Swagger schema to JSON Schema
  */
-export default function convertSchema(schema, root, convertDefinitions = true) {
-  const references = [];
+export function convertSchema(schema, root, copyDefinitions = true) {
```
kylef: Before root was the original Swagger document, instead now it is always the converted JSON Schema definitions, and thus the parameter is renamed to copyDefinitions, to copy the referenced definitions over to the converted schema.

test/json-schema.js
```diff
@@ -310,25 +310,6 @@ describe('Swagger Schema to JSON Schema', () => {
       });
     });
 
-    it('dereferences root reference and converts to JSON Schema', () => {
```
kylef: This test is no longer applicable, since the referenced document is no longer being converted as the source is already converted to JSON Schema. Removing the "converts to JSON Schema" part of the test is already the same test as the test case above so becomes redundant.

test/json-schema.js
```diff
@@ -434,37 +415,7 @@ describe('Swagger Schema to JSON Schema', () => {
       });
     });
 
-    it('copies references to schema and converts them to JSON Schema', () => {
```
kylef: This test is no longer applicable, since the referenced document is no longer being converted as the source is already converted to JSON Schema. Removing the "converts to JSON Schema" part of the test is already the same test as the test case above so becomes redundant.

PR #217, opened 2018-10-20
Handle example references

kylef:
Fixes a regression introduced in 0.22.0 where the parse result may contain invalid references inside a JSON Schema for example values if they used references. This regression also caused `$ref` to be incorrectly present in Data Structure sample values. The `$ref` was often invalid because it referred to `example` which isn't valid in JSON Schema (example vs examples array). The solution is to expand examples.

```yaml
swagger: "2.0"
info:
  version: 1.0.0
  title: Data Structure Generation
paths:
  /user:
    get:
      description: Get a resource
      operationId: getResource
      responses:
        200:
          description: response description
          schema:
            type: object
            example:
              id: 123
              user:
                $ref: '#/definitions/User/example'
            properties:
              id:
                type: number
              name:
                $ref: '#/definitions/User'
definitions:
  User:
    type: object
    example:
      name: Doe
```

### Dependencies

- [x] https://github.com/apiaryio/swagger-zoo/pull/72

### DIFF
--- a/src/json-schema.js
+++ b/src/json-schema.js
@@ -5,7 +5,73 @@ function isExtension(value, key) {
   return _.startsWith(key, 'x-');
 }
 
-function convertSubSchema(schema, references) {
+export function parseReference(reference) {
+  const parts = reference.split('/');
+
+  if (parts[0] !== '#') {
+    throw new Error('Schema reference must start with document root (#)');
+  }
+
+  if (parts[1] !== 'definitions' || parts.length !== 3) {
+    throw new Error('Schema reference must be reference to #/definitions');
+  }
+
+  const id = parts[2];
+
+  return id;
+}
+
+function lookupReference(reference, root) {
+  const parts = reference.split('/').reverse();
+
+  if (parts.pop() !== '#') {
+    throw new Error('Schema reference must start with document root (#)');
+  }
+
+  if (parts.pop() !== 'definitions') {
+    throw new Error('Schema reference must be reference to #/definitions');
+  }
+
+  const id = parts[0];
+  let value = root.definitions;
+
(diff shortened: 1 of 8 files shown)


src/json-schema.js
```diff
@@ -5,7 +5,73 @@ function isExtension(value, key) {
...
+export function parseReference(reference) {
+  const parts = reference.split('/');
+
+  if (parts[0] !== '#') {
+    throw new Error('Schema reference must start with document root (#)');
+  }
+
+  if (parts[1] !== 'definitions' || parts.length !== 3) {
+    throw new Error('Schema reference must be reference to #/definitions');
+  }
```
honzajavorek: One would expect these are handled by the Swagger parser itself

src/json-schema.js
```diff
@@ -5,7 +5,73 @@ function isExtension(value, key) {
...
+function lookupReference(reference, root) {
+  const parts = reference.split('/').reverse();
+
+  if (parts.pop() !== '#') {
+    throw new Error('Schema reference must start with document root (#)');
+  }
+
+  if (parts.pop() !== 'definitions') {
+    throw new Error('Schema reference must be reference to #/definitions');
+  }
```
pksunkara: Probably because I am out of context, but I don't understand why we removed the usage of `parseReference` from this function and duplicated the code with minor difference.

kylef: `parseReference` requires it to be a reference of a definition. Whereas `lookupReference` scans for any arbitary reference inside `definitions`. The latter function is used purely for providing dereferenced values so it isn't strict about the path itself. `lookupReference` is used to copy a reference into the same path so it is very strict about the definition.

PR #226, opened 2018-11-29
fix(json-schema): always pass the swagger argument

opichals:
This is to avoid the following problem where the `convertSubSchema` is called recursively but not passing all of its arguments.

```
TypeError: Cannot read property 'definitions' of undefined
    at lookupReference (/Users/standa/Dev/tmp/node_modules/fury-adapter-swagger/lib/json-schema.js:58:20)
    at convertExample (/Users/standa/Dev/tmp/node_modules/fury-adapter-swagger/lib/json-schema.js:80:19)
    at example.map.value (/Users/standa/Dev/tmp/node_modules/fury-adapter-swagger/lib/json-schema.js:77:33)
    at Array.map (<anonymous>)
    at convertExample (/Users/standa/Dev/tmp/node_modules/fury-adapter-swagger/lib/json-schema.js:77:20)
    at _lodash2.default.forEach (/Users/standa/Dev/tmp/node_modules/fury-adapter-swagger/lib/json-schema.js:87:21)
```

### DIFF
--- a/src/json-schema.js
+++ b/src/json-schema.js
@@ -77,6 +77,8 @@ function convertSubSchema(schema, references, swagger) {
     return { $ref: schema.$ref };
   }
 
+  const recurseConvertSubSchema = s => convertSubSchema(s, references, swagger);
+
   let actualSchema = _.omit(schema, ['discriminator', 'readOnly', 'xml', 'externalDocs', 'example']);
   actualSchema = _.omitBy(actualSchema, isExtension);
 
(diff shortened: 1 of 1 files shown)


kylef:
Thanks @opichals, I've taken this commit and merged it in via #227 with a fix for the problem and an integration test for the particular broken case. This does uncover some untested code paths, I've created https://github.com/apiaryio/fury-adapter-swagger/issues/229 to track adding additional coverage.

PR #228, opened 2018-11-29
Enable schema with references generation test

kylef:
This was fixed in other PRs and the test wasn't updated before.

### DIFF
--- a/test/generator.js
+++ b/test/generator.js
@@ -93,8 +93,7 @@ describe('bodyFromSchema', () => {
     expect(body.length).to.equal(5);
   });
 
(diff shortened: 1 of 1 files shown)


test/generator.js
```diff
@@ -93,8 +93,7 @@ describe('bodyFromSchema', () => {
     expect(body.length).to.equal(5);
   });
 
-  // FIXME: Disabled, see https://github.com/apiaryio/fury-adapter-swagger/pull/212
-  xit('can generate a structure with references', () => {
+  it('can generate a structure with references', () => {
```
pksunkara: For future, you should use `it.skip`

PR #230, opened 2018-12-06
Fix generating data structure from schema from nullable type

kylef:
When a Swagger schema contains a x-nullable it creates a JSON Schema with a type array which was previously unhandled and thus caused `x-nullable` not to work in the data structure generator.

### DIFF
--- a/test/fixtures/data-structure-generation-nullable-member.json
+++ b/test/fixtures/data-structure-generation-nullable-member.json
@@ -1,0 +1,139 @@
+{
+  "element": "parseResult",
+  "content": [
+    {
+      "element": "category",
+      "meta": {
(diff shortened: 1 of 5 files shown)


test/fixtures/data-structure-generation-nullable-member.json
```diff
@@ -0,0 +1,139 @@
...
+                                      }
+                                    ]
+                                  }
+                                },
+                                "content": {
+                                  "key": {
+                                    "element": "string",
+                                    "content": "name"
+                                  },
+                                  "value": {
```
kylef: @ajkl2533 Previously, the property `value` on this `content` (KeyValuePair) was mising.

PR #231, opened 2018-12-06
Generate data structures from Swagger Schema

kylef:
In a previous changeset, I moved DataStructureGenerator to take a JSON Schema instead of the Swagger Schema. This resulted in various problems due to not fully supporting JSON Schema. This change reverts back to the old behaviour of passing a Swagger Schema into DataStructureGenerator and removing some of the handling of JSON Schema specific properties (`examples`) will are not found in Swagger Schema.

Since the provided Swagger Schema may not be dereferenced, we also have to dereference the `example` values where users are putting references in the structures.

As a side effect, this fixes a bug where generating a data structure from a nullable type (`x-nullable`) was being converted to the `type` array which wasn't supported. Thus now that the Swagger Schema is used in DataStructure generator it receives `x-nullable` as expected.

Secondly, the last two commits fix another bug which was with certain API Description documents which had `$ref` in example values to another properties example value would fail. This requied two fixes, the first which fixes a problem where the reference destination was not copied correctly to the schema. The second problem is that even with the valid `$ref` Faker wasn't handling it properly. I tested it out using Faker's async API (`resolve`) which worked but the problem with using their async API is that it would require a lot of refactoring of this Swagger adapter to invoke async code, thus instead I made our `dereference` function handle circular references (by early exiting) an use that before faker. This may have a positive impact on performance of circular reference handling as we have more control about exiting earlier than what Faker is doing.

Superceeds https://github.com/apiaryio/fury-adapter-swagger/pull/230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Fury Swagger Parser Changelog
 
+## Master
+
+### Bug Fixes
+
+- Fixes using `x-nullable` in a schema which previously caused the schema not
+  to be converted to a data structure. For example, the following schema
+  wouldn't result in a data structure:
+
+  ```yaml
+  type: string
+  x-nullable: true
+  ```
+
 ## 0.22.4 (2018-11-29)
 
 ### Bug Fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
   x-nullable: true
   ```
 
+- Fixes some cases where rendering a JSON example from a Swagger Schema using
+  example values which contain references would fail.
+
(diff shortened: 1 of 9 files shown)


char0n:
Thank you for fixing this. What would be the estimated ETA of release to production ?

kylef:
@char0n I'm a little concerned about the performance implications. I'd investigate that further today and see about releasing this.

kylef:
Ok, seems after deeper investigation that it isn't really a problem. I have a single Swagger document that already hit our timeouts that got worse, but for the other large Swagger documents I tested I did not notice any real difference in performance.

PR #232, opened 2018-12-07
Support references in allOf for objects (object inheritance / mixins)

kylef:
This fixes cases of using `allOf` in conjunction with a data structure.

For example, the following will create an object which inherits from `User`:

  ```yaml
  allOf:
    - $ref: '#/definitions/User'
    - properties:
        id:
          type: string
  ```

Secondly, when you reference multiple objects using `allOf` they will be treated as mixins:

  ```yaml
  allOf:
    - $ref: '#/definitions/BaseUser'
    - $ref: '#/definitions/UserMixin'
  ```

Secondly, I've found that a lot of people miss putting `type: object` and we can try to infer the type from the presence of `properties`. I found many cases where users forget this and that causes no data structure to be rendered thus breaking a reference.

### DIFF
--- a/test/fixtures/circular-example.json
+++ b/test/fixtures/circular-example.json
@@ -407,6 +407,35 @@
                       "key": {
                         "element": "string",
                         "content": "company"
+                      },
+                      "value": {
+                        "element": "object",
+                        "content": [
+                          {
+                            "element": "member",
+                            "attributes": {
+                              "typeAttributes": {
+                                "element": "array",
+                                "content": [
+                                  {
+                                    "element": "string",
+                                    "content": "optional"
+                                  }
+                                ]
+                              }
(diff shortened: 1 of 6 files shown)


test/fixtures/circular-example.json
```diff
@@ -407,6 +407,35 @@
                       "key": {
                         "element": "string",
                         "content": "company"
+                      },
```
kylef: We forgot to put `type: object` in one of our test fixtures too.

kylef:
There is also some other cases that are not currently handled, I don't think they are as critical though so added for our backlog:

- Error when `allOf` is used with contradictory `type` https://github.com/apiaryio/fury-adapter-swagger/issues/233

